PR #24, opened 2017-12-27
Fix: currentNodeIsForward 

6br:
Hello, I found a bug that if the nodeId is assigned as `0` by `generateNodeMap`, then it cannot distinguish `x = 0` or `x = -0` by `x >= 0` in `currentNodeIsForward` variable. It caused the unexpected output in an inversion region.


wolfib:
Good catch, I did not know about -0.

Thanks a lot!

PR #33, opened 2018-02-01
Add gbwt

adamnovak:
This is on top of #31.

I don't have a good setup to test the Docker changes at the moment, but this seems to work fine outside the container.

Maybe there should be container-building CI.


adamnovak:
I think we really need to wait for https://github.com/vgteam/vg/pull/1418 to land in a vg docker for this to be ready; GBWT haplotype traces are wrong without it.

Also, adding the extra dropdown to pick a GBWT file budges the path select dropdown onto the next line on my oridnary desktop screen, so that's probably bad. I'm not sure we need to be able to use more than one GBWT with an xg, since because of the way we build them (i.e. together in the same `vg index` command) you have either exactly one sensible GBWT file for an xg file or none at all.

wolfib:
I had a look at this and in general I think it looks great.

Still, I would want to test this before merging, so I'll wait for the changes to vg. Especially the docker stuff will be interesting. I believe that the docker base image is a minimal linux distro without bash, which might require some changes to the scripts.

PR #53, opened 2018-05-23
 Update checkbox states across page refreshes #34

### LINKED ISSUE #34: Checkbox states are retained across page refreshes by the browser and the frontend doesn't notice
In Firefox at least, when you refresh the page, the checkboxes and select dropdowns retain their previous states. For example, if you turned off `Remove redundant nodes `, the checkbox will be clear on page refresh.

But the frontend doesn't know about this; it will render things with redundant nodes removed until you re-check and re-un-check the checkbox.

Either the checkboxes and dropdowns should be reset by the page to their default values on page load, or some logic should be added to see if the browser has adjusted the state of these controls, so their settings are respected.

tanessav:
This fixes #34. When the user updates the browser, the checkboxes go back to their default state.

### DIFF
--- a/frontend/app/index.html
+++ b/frontend/app/index.html
@@ -82,7 +82,6 @@
       </svg>
     </div>
 
-<!--         <script id="template" type="text/ractive"> -->
     <div class="container">
       <div id="accordion" role="tablist">
         <div class="card">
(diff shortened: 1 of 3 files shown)


adamnovak:
Sorry I took so long to have a look at this. This looks good, except for the Ractive dependency that is still being added. We ought to pull that line out of package.json before merging.

Issue #103, opened 2021-03-01
We drop the first character of nodeName even when it is significant


adamnovak:
Here we expect to be dropping a minus sign, but we also can end up dropping a digit.

https://github.com/vgteam/sequenceTubeMap/blob/f61437b7ea56ca678b36f3736a4357f2480ea9c0/src/util/tubemap.js#L962

Issue #93, opened 2020-08-21
generateTrackIndexSequences returns negative values when there are no negative array members


Nheyer:
https://github.com/vgteam/sequenceTubeMap/blob/713b2d993099c4b11326d45b18cf87a4ce0a8022/src/util/tubemap.js#L963

I have spent the last few dayes trying to figure out why when I get to a specific part of the VG graph that was generated from cactus, and simplified via VG simplify, sequenceTubeMap yealds a blank screen and crashes, it seems that this code here is designed to map the absolute value of the nodemap_value , assuming that a negative key ( that is a key starting with the char '-')  implies a negative node index in the map, but this is NOT always the case, and such a case causes an undefined object to be generated when exicuting [this](https://github.com/vgteam/sequenceTubeMap/blob/713b2d993099c4b11326d45b18cf87a4ce0a8022/src/util/tubemap.js#L420)
when we get to [here](https://github.com/vgteam/sequenceTubeMap/blob/713b2d993099c4b11326d45b18cf87a4ce0a8022/src/util/tubemap.js#L2923) and this leads to an exception being thrown [here](https://github.com/vgteam/sequenceTubeMap/blob/713b2d993099c4b11326d45b18cf87a4ce0a8022/src/util/tubemap.js#L2926) causing rendering to be halted indefinitely.

To sum up I just need to know if having negative node-indexes anywhere is an expected behavior, and if changing the code here will cause  any runtime errors, given that  when I modify it locally everything seems to work

Nheyer:
@adamnovak

adamnovak:
There should be no real zero or negative node IDs in the underlying graph.

However, it does seem like we're using negation of a node ID to mean something, and I have no clue what. Maybe strand?

Nheyer:
Do you have an idea who would, because if so, my simple absolute value may be breaking things, even though it seems to be working.

adamnovak:
@wolfib should know, but I think he's moved on to other projects. Maybe @glennhickey? He apparently last touched a bunch of the vg-side code that talks to the tube map.

wolfib:
I am not sure about the context here, but here is what I remember in general: nodes themselves don't have negative indices, but the sequence of nodes a track passes through can contain negative node indices. This simply means that the track traverses the node with that index in reversed direction (i.e. from right to left in the resulting graph).

Nheyer:
ahh ok @wolfib  if this is what is going on pr #94  should fix the issue I was having properly

PR #110, opened 2021-03-01
Regularize orientation

adamnovak:
This makes node name orientation be handled via functions, instead of strign
indexing and substr.

It also makes node index orientation be removed (with Math.abs) everywhere an
indexSequence is read.

It partially reverts f61437b7ea56ca678b36f3736a4357f2480ea9c0, and instead of
making indexSequence always be positive numbers, we keep tracking orientation
through the index sequences and instead fix `drawRuler` and similar functions
to pull out the non-oriented node ID when looking for the node. This should fix
up #93 in a different way.

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -133,7 +133,7 @@ let bed;
 // main function to call from outside
 // which starts the process of creating a tube map visualization
 export function create(params) {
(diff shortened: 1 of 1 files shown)


adamnovak:
Since we no longer fiddle around with node names in an uncontrolled way, this will fix #103.

PR #122, opened 2021-03-09
Handle legend properly for reverse strand paths

### LINKED ISSUE #88: confusion with genomic positions (coordinate axis) of minus strand genes
I'm analysing a bacterial mini pangenome with five published strains and installed IVG to be able to visualise the graph. Some of the strains have a major inversion of about half of the genome compared to others, so genes are on the plus strand in some genomes and on the minus in others. In IVG I can enter a genome position from my selected reference strain and see e.g. the start of a gene of interest and then scroll to the right to see the gene sequence - if it is on the plus strand this makes perfect sense.

If the gene is located on the minus strand, I can enter the nucleotide position of the start of the gene and the right region will be displayed. Due to the orientation of the graph, I can again scroll right to follow the gene sequence, **but** the nucleotide positions shown on the axis will increase. This is confusing, since the official genomic position of the end of the gene is smaller than the position of the start, and I can use this smaller number to find the end of the gene in the graph.

Example: the annotation is "1043983 - 1045314 (- strand)". If I search for position 1045314 I find the start of the gene, and then I scroll to the right, e.g. to 1045320 to follow the gene. If I want to jump to the end of the gene without scrolling, I have to enter 1043983, though.

I have an example with pictures [here](https://cpang.netlify.app/post/day-3-bacterial-pangenomics/#visualisation-with-ivg).

adamnovak:
The legend display is wrong when a path visits all the nodes in their local reverse orientations. This should fix that eventually.

Closes #88.


adamnovak:
This should work now.

I've rewritten the tick generator in `drawRuler()` to count along the path and properly place ticks at the right places when the path visits a node in reverse. Instead of trying to walk in X coordinate and path position together to prevent ticks from crowding each other, now I sort the ticks by X coordinate and drop enough of them to ensure good separation.

I've also stopped completely reversing paths where all the nodes they visit are in reverse; I couldn't figure out why that was done, dropping it doesn't break any tests, and it was getting in the way of knowing what orientation nodes were in inside the tick generator.

adamnovak:
image

PR #140, opened 2021-10-11
Interface change, pre-fetched chunks, and small fixes

jmonlong:
## Interface change

This PR changes a bit the interface to be simpler and faster to use when having to query many regions. There is now one "Region" search field where the user can input coordinates, node IDs. It removes the "Length" and "Unit" field.

We can now just paste a region definition like in the UCSC Genome Browser (`CHR:START-END`) to the search field and press Enter.

We can still use a position+length query by using the `+` symbol. For example, `chr1:1000+100` does the same as before for `1000` in the "Start" field and `100` in the "Length" field.

Node IDs are queried by using `node` for sequence names, for example `node:100-110` or `node:100+10`.

I added a help popup text next to the search/region field explaining those accepted queries.

Of note, there is still a scrolldown menu for the path names, but it's not mandatory to use it. It's useful to know the path names in the graph. Selecting a path there will prefix the search field with the selected sequence name.

## Pre-fetched chunks

To save time when there are a lot of regions to visualize, we can now pre-fetch the chunks. Then tubemap can skip the fetching and show these regions fast.

The subgraphs need to be pre-fetched using `vg chunk` like shown in `prepare_chunks.sh` (in *scripts*).
Then compile those regions in a BED file with two additional columns: 

- a description of the region (column 4)
- the path to the output directory of the chunk (column 5). 

There is now an example in `cactus.bed` (in *exampleData*). 

This BED file will be read if placed in the `dataPath` directory, like for other files to mount. For other data sources, it can also be specified in the `config.json` like other files.

In the interface, a new scrolldown menu will show the region descriptions and fill the Search/Region field when one is selected.

Screenshot from 2021-10-11 10-41-53

## Small fixes and additions

- Fix the landing page to be the first data source defined in the `config.json` instead of a hard-coded example. It will also load the appropriate files (e.g. BED/regions) but won't show a graph. For that, the user has to press the 'Go' button.
- Fix mechanism to flip nodes that are traverse most of the time in reverse (by paths), see example below.
- Use one color exclusively for the first path, plus a new palette with one purple color and shades of grey. This makes it easier to spot the reference path among lots of haplotypes, for example.
- The queried region is (more or less?) highlighted by two yellow circles on the ruler.
- Pressing `Enter` while on the search/position field launches the query (like pressing Go) instead of restarting the app.
- Hovering over a path shows its name, and over a read the read name.
- Button to select/deselect all paths. Makes it easier to show just the reference and a few haplotypes when there are a lot of haplotypes.

Example of node flipping correctly now compared to before

Screenshot from 2021-09-28 10-18-33

Screenshot from 2021-09-28 10-22-25


src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
-  if (req.withGam) {
-    // Use a GAM index
-    vgChunkParams.push('-a', `${dataPath}${gamFile}`, '-g');
+  // parse region
+  // either a range -> seq:start-end
+  // or a position and distance -> seq:post+distance
+  let region = req.body.region;
+  let region_col = region.split(":");
+  let start_end = region_col[1].split("-");
+  let pos_dist = region_col[1].split("+");
```
adamnovak, line 183: This is just going to throw an error on the server if the client's requested region doesn't have ":". We need to actually check for it I think.

src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
+      nb_regions_in_bed = req.body.regionInfo['desc'].length;
+    }
+    while (i < nb_regions_in_bed && chunkPath === ''){
+      let region_chr = req.body.regionInfo['chr'][i];
+      let region_start = req.body.regionInfo['start'][i];
+      let region_end = req.body.regionInfo['end'][i];
+      if(region_chr.concat(':', region_start, '-', region_end) === region){
+	if(req.body.regionInfo['chunk'][i] !== ''){
+	  chunkPath = req.body.regionInfo['chunk'][i]
+	}
```
adamnovak, line 210: We need to handle missing/wrong-type structures in `req.body.regionInfo`.

src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
+      }
+      i += 1;
+    }
+    // check that the 'chunk.vg' file exists in the chunk folder
+    let chunk_file = `${dataPath}${chunkPath}/chunk.vg`;
```
adamnovak, line 215: `chunkPath` has been sent in from the client; we definitely can't just paste it into a file path without making sure it doesn't go up a directory.

src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
-  if (
-    Object.prototype.hasOwnProperty.call(req.body, 'byNode') &&
-    req.body.byNode === 'true'
-  ) {
-    vgChunkParams.push('-r', position, '-c', distance);
-  } else {
+  
+  if(chunkPath === ''){
+    // call 'vg chunk' to generate graph
+    let vgChunkParams = ['chunk', '-x', `${dataPath}${xgFile}`];
```
adamnovak, line 226: We might be missing some path sanitization here too.

src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
+    if (req.withGam) {
+      // Use a GAM index
+      vgChunkParams.push('-a', `${dataPath}${gamFile}`, '-g');
```
adamnovak, line 229: And we need to sanitize the GAM paths.

src/server.js
```diff
@@ -170,103 +174,195 @@ api.post('/getChunkedData', (req, res) => {
...
+    }
+    if (req.withGbwt) {
+      // Use a GBWT haplotype database
+      vgChunkParams.push('--gbwt-name', `${dataPath}${gbwtFile}`);
```
adamnovak, line 233: And we need to sanitize the GBWT paths.

src/server.js
```diff
@@ -452,6 +555,47 @@ api.post('/getPathNames', (req, res) => {
...
+    bedRegions: []
+  };
+
+  let bed_info = {chr:[], start:[], end:[], desc:[], chunk:[]};
+
+  if(req.body.bedFile != 'none'){
+    const bedFile =
+	  req.body.isUploadedFile === 'true'
+	  ? `./${req.body.bedFile}`
+	  : `${MOUNTED_DATA_PATH}${req.body.bedFile}`;
```
adamnovak, line 570: We're just letting everybody on the Internet read every TSV on the system here, such as for example `/etc/shadow`; we need to make sure that this file is actually in the directory it is supposed to be in when we read it.

adamnovak (changes requested):
This looks mostly good, but there are some security holes we probably want to patch before we actually merge this. In general it looks like we haven't been making sure that files are actually where they're supposed to be when we go to read them, and that gets worse when we start working with plain text files instead of running everything through `vg`.

Also some of the indentation is weird.

Also, it looks like with the new "Data" dropdown we always use e.g. one specific GAM with each xg, right? Previously, we could toggle between GAMs or GBWTs if we had several of them. Can we not do that anymore?

I've started a little on the indents, but I haven't gotten to the security stuff yet. I can do it eventually if you don't want to @jmonlong.

src/server.js
jmonlong: Good point, I've added checks in the latest comment. I've also added a way to show an error message on the front end for multiple errors, including this one.

src/server.js
jmonlong: `req.body.regionInfo` should have the proper structure by construction (or be empty) by the `getBedRegions` function in `server.js`

src/server.js
jmonlong: Commit 8db286 now checks that the paths used here and in general are "allowed" which I defined as no backtracking `..` and not starting with `/etc`, `/sys`, `/usr`.

src/server.js
jmonlong: https://github.com/vgteam/sequenceTubeMap/pull/140/commits/8db286b9f20863a16979c82155aa6532bd04a9ae adds sanitizing of the paths by checking that the path is allowed and the files have the expected extension. Here that `xgFile` ends with `.xg`

src/server.js
jmonlong: https://github.com/vgteam/sequenceTubeMap/pull/140/commits/8db286b9f20863a16979c82155aa6532bd04a9ae commit check and sanitize this path too

src/server.js
jmonlong: https://github.com/vgteam/sequenceTubeMap/pull/140/commits/8db286b9f20863a16979c82155aa6532bd04a9ae commit check and sanitize this path too

src/server.js
jmonlong: It couldn't really read those kind of files because of permission issues (I tried). Anyway, [now](https://github.com/vgteam/sequenceTubeMap/pull/140/commits/8db286b9f20863a16979c82155aa6532bd04a9ae) there are checks to make sure the files have the proper extensions (in this case `.bed`). And also the path can have `..` or `/etc`, `/sys`, `/usr` in them, just in case.

jmonlong:
It's now handling the paths and files much more safely: checking the expected extension of each file before doing anything (`.bed`, `.gam`, `.gbwt`, `.xg`) and also raising an error if a path looks unsafe (for now unsafe means containing a `..` or starting with `/etc`, `/usr`, `/sys`).

I've added a new error mechanism for the Header to display those kinds of file/path/config.json errors. 

I also fixed some small issues I encountered:
- BED file or paths didn't load on the landing page if not on the internal path.
- Getting paths from .xg worked only for the mounted files. I've separated the `exampleData` folders to help testing when the data path and internal data path are different.
- Make sure the path are glued properly, avoiding things like `path_to_chunk//chunk.vg`

src/server.js
adamnovak: I'd prefer not to store server-generated and user-supplied stuff next to each other in the same object, since it's easy then to forget the provenance or trustworthiness of a field. But I suppose it won't do more than throw a server error if we're wrong here.

src/server.js
adamnovak: I don't think just banning `/etc`, `/sys`, and `/usr` is right. I think the user expectation here is that the tube map server is only going to serve stuff out of the specific configured backend directory/directories, not that it's going to potentially serve everything on the hard drive except those. What if the tube map is supposed to be serving `/home/jmonlong/publicData/tubeMap/HPRC.gbz` and we definitely don't want it to be able to serve `/home/jmonling/privateData/peopleWithCancer/BrettThePatientsEntireGenome.gbz`?

I think we need a collection of the allowed directories, and we need to verify that the path we are looking at is actually in one of them.

src/server.js
```diff
@@ -380,40 +537,89 @@ function processRegionFile(req, res) {
...
 
+// check that the path is absolute and not pointing at something it shouldn't
+function isAllowedPath(inputPath) {
+  let allowed_path = true;
+  let forbidden_root = ['/etc', '/usr', '/sys'];
+  for (var i = 0; i < forbidden_root.length; i++){
+    if(inputPath.startsWith(forbidden_root[i])){
+      allowed_path = false;
+    }
+  }
```
adamnovak, line 566: This also doesn't quite work; `/./etc/passwd` is still going to pass.

adamnovak:
OK, I revised the input validation and error checking to be more the way I want it.

@jmonlong Have you tried going straight to one of the synthetic data examples? When I did it, it would crash because `tracks` in the tube map module wasn't filled in yet. I added some checks to make sure we don't try to draw the legend without the tracks, but it still doesn't actually work unless you draw a different tube map first. Have you been seeing this or did I cause this with my hacking today?

adamnovak:
It looks like the problem is the new `headerFetched` flag this PR adds; without it, there won't actually be a tube map displayed, and therefore the example data buttons won't render anything.

adamnovak:
Another change I've noticed, once I set it to show the initial tube map again, is that we're parsing the BED from the server into `reginInfo`, and then uploading that as part of our `fetchParams`. So when the BED comes in and `regionInfo` is populated, after the initial render, it means that hitting the Go button is going to redo a bunch of requests, because now it thinks it has to send `regionInfo` back.

I don't think it makes any sense to be sending the parsed `regionInfo` back up to the server, especially when it's full of info not about the region we are trying to look at.

Also, I think I was right the first time in https://github.com/vgteam/sequenceTubeMap/pull/140#discussion_r734084542; the `regionInfo` in the request is coming from the client, and can't be trusted to have the right structure when we look at it. Even if the client is *supposed* to be sending back stuff it retrieved from the server, it might not actually be doing so, and we can't trust that it is.

adamnovak:
Also, when we use a cached chunk, we set `req.tempDir` to the chunk directory. But when a request errors out, we delete `req.tempDir`.

adamnovak (approved):

adamnovak:
OK, I've changed the BED region info to be parsed from the BED on the server instead of shipped back from the client to the server with every request, which let me remove it from the visualization component's fetch parameter state, where it was forcing spurious re-renders by not properly being part of the client-side state of what we're supposed to be rendering.

I've also consolidated the temp directory cleanup so we can't remove any cached chunks. And I've fixed going straight to the toy examples not showing anything/crashing (by ripping out headerFetched), and added a test for those examples so they shouldn't break the same way again.

@jmonlong I think I'm going to merge this. Do you really need the tube map to not draw anything on initial load? Because I think it's much more user-friendly to show the user something by default, instead of just the controls.

PR #147, opened 2022-02-23
Update dependencies

olivewong:
- Updated old dependencies
- Hopefully silence dependabot
- Added package-lock.json
- Tested manually (node v14.17.6) by checking out default data and inputting regions
Screen Shot 2022-02-23 at 12 45 43 PM


adamnovak (approved):
This looks good!

Were the form fields always that wide, like they are in your screenshot, or did they change in this PR? Because I think that form is supposed to be just one line.

olivewong:
> This looks good!
> 
> Were the form fields always that wide, like they are in your screenshot, or did they change in this PR? Because I think that form is supposed to be just one line.

@adamnovak  This is what master looks like for me right now: 
Screen Shot 2022-02-28 at 11 23 47 AM

So I believe it was like that before this PR, but I can look into making the fields smaller like the existing demo site.

adamnovak:
I broke off the wideness into a new issue at #149. We can rearrange them at some point later.

I'm going to go ahead and merge this.

PR #154, opened 2022-04-04
Handle new http-proxy-middleware version

adamnovak:
This changes the proxy setup code to work right with the http-proxy-middleware
that exports multiple functions.

It also makes sure to stop the process if proxy setup fails, because
react-scripts will claim to successfully start the server but not actually open
the port, if proxy setup throws.


src/server.js
```diff
@@ -25,7 +28,7 @@ const INTERNAL_DATA_PATH = config.internalDataPath;
 const UPLOAD_DATA_PATH = "uploads/";
 // This is where we will store per-request generated files
 const SCRATCH_DATA_PATH = "tmp/";
-const SERVER_PORT = config.serverPort || 3000;
+const SERVER_PORT = process.env.SERVER_PORT || 3000;
```
adamnovak, line 31:
```suggestion
const SERVER_PORT = process.env.SERVER_PORT || config.serverPort || 3000;
```

PR #160, opened 2022-05-17
Add link to the current view

olivewong:
### Changes:
- **Adds a 'Copy Link' button that allows the user to link to the current view**
- Refactor the fetchParams and similar data types into a viewTarget 
- Minor fixes in errors and button styling

CopyLink

### Testing:
- E2E tests for copy button before and after hitting 'Go'
- Manually tested on examples (small and cactus), mounted files

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -108,7 +108,7 @@ class App extends Component {
           setColorSetting={this.setColorSetting}
           dataOrigin={this.state.dataOrigin}
           apiUrl={this.props.apiUrl}
-          urlParams={urlParamsToViewTarget()} 
+          viewTarget={urlParamsToViewTarget()} 
         />
         <TubeMapContainer
           viewTarget={this.state.viewTarget}
--- a/src/App.js
+++ b/src/App.js
@@ -81,7 +81,7 @@ class App extends Component {
 
     if (
       !isEqual(this.state.viewTarget, newViewTarget) ||
-      this.state.dataOrigin != dataOriginTypes.API
+      this.state.dataOrigin !== dataOriginTypes.API
     ) {
       this.setState({
         viewTarget: newViewTarget,
@@ -133,7 +133,7 @@ class App extends Component {
           setColorSetting={this.setColorSetting}
           dataOrigin={this.state.dataOrigin}
           apiUrl={this.props.apiUrl}
-          viewTarget={urlParamsToViewTarget(document.location)}
+          defaultViewTarget={urlParamsToViewTarget(document.location)}
           getCurrentViewTarget={this.getCurrentViewTarget}
         />
         <TubeMapContainer
--- a/src/App.js
+++ b/src/App.js
@@ -73,6 +73,7 @@ class App extends Component {
    * @param {ViewTarget} viewTarget - The default or selected data to view
    */
   setCurrentViewTarget = (viewTarget) => {
+    // Remove undefined for equality check
     const newViewTarget = this.removeUndefined(viewTarget);
 
     if (
@@ -86,7 +87,7 @@ class App extends Component {
     }
   };
   getCurrentViewTarget = () => {
-    return this.state.viewTarget;
+    return this.removeUndefined(this.state.viewTarget);
   };
 
   toggleVisOptionFlag = (flagName) => {
--- a/src/App.js
+++ b/src/App.js
@@ -4,17 +4,12 @@ import PropTypes from "prop-types";
 import "./App.css";
 import HeaderForm from "./components/HeaderForm";
 import TubeMapContainer from "./components/TubeMapContainer";
-import { CopyLink, urlParamsToObject } from "./components/CopyLink";
+import { CopyLink, urlParamsToViewTarget } from "./components/CopyLink";
 import CustomizationAccordion from "./components/CustomizationAccordion";
 import { dataOriginTypes } from "./enums";
 import * as tubeMap from "./util/tubemap";
 import config from "./config.json";
 // Data path  - see server.js and config.json
-const dataPaths = {
-  MOUNTED: "mounted",
-  DEFAULT: "default", 
-  UPLOAD: "upload"
-}
 
 class App extends Component {
   constructor(props) {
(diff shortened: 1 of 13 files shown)


src/App.js
```diff
@@ -1,64 +1,42 @@
...
+
+import "./App.css";
+import HeaderForm from "./components/HeaderForm";
+import TubeMapContainer from "./components/TubeMapContainer";
+import { urlParamsToViewTarget } from "./components/CopyLink";
+import CustomizationAccordion from "./components/CustomizationAccordion";
+import { dataOriginTypes } from "./enums";
+import * as tubeMap from "./util/tubemap";
+import config from "./config.json";
+// Data path  - see server.js and config.json
```
adamnovak, line 13: What does this mean?

src/App.js
```diff
@@ -1,64 +1,42 @@
...
-        bedFile: bedFile,
-        // This is the type of data paths we are working with, such as "mounted".
-        // All the paths are scoped to a type on the server side.
-        dataPath: dataPath
-      },
       // This is a little like dataPath, but lets us toggle between data from
       // the server and local test data. TODO: Unify?
-      dataOrigin: dataOriginTypes.API,
+      viewTarget: ds,
+      dataOrigin: dataOriginTypes.API, // EXAMPLE[N] etc.
```
adamnovak, line 26: I guess the comment is meant to explain that you can set dataOrigin to the values for the built-in synthetic examples, right? Could you maybe explain that a bit more here? I don't think that this comment will make sense to other people.

src/App.js
```diff
@@ -1,64 +1,42 @@
...
-        xgFile: xgFile,
-        gbwtFile: gbwtFile,
-        gamFile: gamFile,
-        bedFile: bedFile,
-        // This is the type of data paths we are working with, such as "mounted".
-        // All the paths are scoped to a type on the server side.
-        dataPath: dataPath
-      },
       // This is a little like dataPath, but lets us toggle between data from
       // the server and local test data. TODO: Unify?
```
adamnovak, line 24: This should apply to `dataOrigin` and not `viewTarget`, right? Also, maybe now it needs to say what it means by `dataPath` (i.e. that it's part of `viewTarget` here?).

src/App.js
```diff
@@ -71,65 +49,90 @@ class App extends Component {
...
+   * See https://stackoverflow.com/questions/286141/remove-blank-attributes-from-an-object-in-javascript/38340730#38340730
+   */
+  removeUndefined = (obj) => {
+    return Object.fromEntries(
+      Object.entries(obj).filter(([_, v]) => v != null)
+    );
+  };
+
+  /**
+   * @param {ViewTarget} viewTarget - The default or selected data to view
```
adamnovak, line 71: This doc comment should also explain how the "current" view target is the one displayed in the tube map itself.

src/Types.ts
```diff
@@ -0,0 +1,26 @@
+// Type file for documentation
+// (not actually enforced by Typescript)
+
+type DataPath = "mounted" | "default" | "upload";
+// All the paths are scoped to a type on the server side.
+// This is the type of data paths we are working with, such as "mounted", "upload", or "default" - see server.js.
```
adamnovak, line 6: This ought to be above what it is documenting.

src/Types.ts
```diff
@@ -0,0 +1,26 @@
...
+// These parameters can be extracted from defaults (config), Header Form input, or URL parameters
+type ViewTarget = {
+  region: string; // Format: path:start-end
+  xgFile: string;
+  gbwtFile?: string;
+  gamFile?: string;
+  bedFile?: string;
+// Non-essential to server, used for examples
+
+  name?: string;
```
adamnovak, line 22: Shouldn't the blank line be above the comment here?

src/components/HeaderForm.js
```diff
@@ -1,134 +1,143 @@
...
   initState = () => {
-    let ds = DATA_SOURCES[0];
-    const xgSelect = ds.xgFile ? ds.xgFile : 'none';
-    const bedSelect = ds.bedFile ? ds.bedFile : 'none';
-    const dataPath = ds.useMountedPath ? 'mounted' : 'default';
-
-    this.setState(state => {
-      if (bedSelect !== 'none'){
+    // Populate
+    // TODO: same issue with different types as in App
```
adamnovak, line 74: Is this TODO not fixed now that everything is a ViewTarget?

src/components/HeaderForm.js
```diff
@@ -1,134 +1,143 @@
...
 
   componentDidMount() {
     this.initState();
     this.getMountedFilenames();
     this.setUpWebsocket();
   }
+  DATA_NAMES = DATA_SOURCES.map((source) => source.name);
 
-  // init with the first data source
+  // init with the URL params, or first data source if none
```
adamnovak, line 71: The part about URL params isn't true; we use `viewTarget` from the props.

src/components/HeaderForm.js
```diff
@@ -540,7 +575,8 @@ HeaderForm.propTypes = {
   dataOrigin: PropTypes.string.isRequired,
   setColorSetting: PropTypes.func.isRequired,
   setDataOrigin: PropTypes.func.isRequired,
-  setFetchParams: PropTypes.func.isRequired
+  setCurrentViewTarget: PropTypes.func.isRequired,
+  viewTarget: PropTypes.any, // Header Form State, may be null if no params in URL. see Types.ts
```
adamnovak, line 579: Since there's already code in App to pick a default ViewTarget if one isn't in the URL parameters, maybe this should be a required prop, so we don't need to look in the config inside this component as well?

It also might make sense to call it `defaultViewTarget` instead of `viewTarget`, because it's yet a third ViewTarget floating around, in addition to the "current" and "next" ones, and they need to be kept straight.

src/components/TubeMapContainer.js
```diff
@@ -28,7 +28,7 @@ class TubeMapContainer extends Component {
         ? this.getRemoteTubeMapData()
         : this.getExampleData();
     } else {
-      if (JSON.stringify(this.props.fetchParams) !== JSON.stringify(prevProps.fetchParams)) {
+      if (JSON.stringify(this.props.viewTarget) !== JSON.stringify(prevProps.viewTarget)) {
```
adamnovak, line 31: Since we now have `react-fast-compare` in the project, we should use that here instead of JSON stringification.

src/App.js
```diff
@@ -1,64 +1,42 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
-import './App.css';
-import HeaderForm from './components/HeaderForm';
-import TubeMapContainer from './components/TubeMapContainer';
-import CustomizationAccordion from './components/CustomizationAccordion';
-import { dataOriginTypes } from './enums';
-import * as tubeMap from './util/tubemap';
-import config from './config.json';
+import React, { Component } from "react";
```
adamnovak, line 1: It looks like you've reformatted everything automatically to use double quotes, whereas previously the convention in the codebase was mostly single quotes.

Are double quotes really better? Are they the default of the popular formatter you're using and so we should switch because that's what popular formatters do?

adamnovak (changes requested):
This looks pretty good! I found a few comments that I think should be elaborated and some places where we might want to change a variable name or more consistently use a new dependency, which I think we ought to do. But after that we can merge this.

src/App.js
olivewong: They are the default of Prettier, the formatter I'm using, which just automatically added double quotes. I don't have any strong opinion on this, and after Googling it seems there's not a consistent rule. If you'd like I can revert to single quotes?

Issue #158, opened 2022-05-16
File upload form has extra BED-file-related controls that shouldn't be there


adamnovak:
On the "custom (file upload)" view, the "Path name" and "Region" dropdowns show up when they should not. I think these are related to @jmonlong's BED file support changes.

Here's a screenshot where I've drawn a little bracket next to the offending controls:
![image](https://user-images.githubusercontent.com/5062495/168674447-d3313c9d-ce19-45b9-a942-ae5628ec118c.png)

To reproduce:

1. Go to https://vgteam.github.io/sequenceTubeMap/
2. Set "Data" to "custom (file upload)"
3. "Path name" dropdown shows but should not; there's no sensible contents for it that the server could have.
4. Same for "Region" dropdown.

PR #161, opened 2022-05-26
Fix issue #158 Hide path and region select when 'Data: custom file upload' is selected

olivewong:
Addresses issue #158 by hiding the region select dropdown and the path option (which comes from a BED file but there is currently no way to upload a BED file). 

After:
Screen Shot 2022-05-26 at 2 23 48 PM

Before:
image


src/components/HeaderForm.js
```diff
@@ -21,46 +21,46 @@ const dataTypes = {
...
+  regionSelectOptions: ["none"],
+  // This tracks several arrays of BED region data, stored by data type, with
+  // one entry in each array per region.
+  regionInfo: {},
+  regionSelect: "none",
+
+  pathSelectOptions: ["none"],
+  pathSelect: "none",
+
+  xgFile: undefined,
```
adamnovak, line 46: At some point we should add some comments here avout what all these fields mean and why we have e.g. both an `xgSelect` and an `xgFile`.

src/components/HeaderForm.js
```diff
@@ -270,15 +270,10 @@ class HeaderForm extends Component {
...
-        return {
-          xgFile: state.xgSelect,
-          gbwtFile: state.gbwtSelect,
-          gamFile: state.gamSelect,
-          bedFile: state.bedSelect,
-          dataPath: "upload",
-          dataType: dataTypes.FILE_UPLOAD,
-        };
+      this.setState({
+        ...EMPTY_STATE,
```
adamnovak, line 274: Won't this clear out any `error` that has been set? Is that likely to be a problem?

adamnovak (changes requested):
I think this looks good, but it wants to be rebased onto current master, so that all the commits from the last PR don't show up in what's being PR'd.

adamnovak:
Might be easier to cherry-pick than to do a `git rebase -i`, since there's only 2 commits we actually need out of the 24 that are here.

PR #168, opened 2022-06-09
Clear region whenever switching between data types

### LINKED ISSUE #166: Region/path not being cleared when switching to file upload or mounted file mode
To reproduce:

1. Select an example (i.e. BRCA)
2. Select custom (mounted files) or custom (file upload)
3. See that region is not cleared in the form even though it may be in state
Screen Shot 2022-06-06 at 12 09 12 PM

Solution:
- Clear both region and path when switching data types

olivewong:
This solves #166 by setting region to 'none' whenever a data type is switched
Clear region

### DIFF
--- a/src/components/HeaderForm.js
+++ b/src/components/HeaderForm.js
@@ -52,7 +52,7 @@ const EMPTY_STATE = {
   gamFile: undefined,
   bedFile: undefined,
   dataPath: undefined,
-  region: undefined,
+  region: "none",
   name: undefined,
 
   dataType: dataTypes.BUILT_IN,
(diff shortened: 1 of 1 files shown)


src/components/HeaderForm.js
```diff
@@ -52,7 +52,7 @@ const EMPTY_STATE = {
   gamFile: undefined,
   bedFile: undefined,
   dataPath: undefined,
-  region: undefined,
+  region: "none",
```
adamnovak, line 55: Does this have to be "none" and not, say, an empty string?

Why does this part help solve the problem? Or was it adding `...EMPTY_STATE,` to the code to swap over to mounted files that did it? Or a combination of the two?

adamnovak (commented):
Looks like this automatically came along with the other PR. I'm still suspicious of the justification for some stuff here, but it seems to actually work in practice.

PR #172, opened 2022-07-02
Add autocomplete field to replace multiple duplicate fields ('Region:', 'Path:', and 'Region:')

olivewong:
### Changes
- Replace double region and path inputs with one autocomplete input
- Show Autocomplete suggestions to choose from
- Generate autocomplete suggestions from bed files

### Motivation
- Provide clarity to UI: previously, the duplicate region fields, with one containing duplicate information from the path field could be confusing and redundant to the user
- Save space

### Screenshots
After:
Autocomplete-gif

Screen Shot 2022-07-18 at 4 43 39 PM

Before:
Screen Shot 2022-07-18 at 4 50 25 PM

### Testing

Manual and E2E testing in `end-to-end.js` and `App.test.js`


src/App.test.js
```diff
@@ -21,17 +25,18 @@ it("allows the data source to be changed", () => {
 
 it("allows the start to be cleared", async () => {
   render(<App />);
-  expect(screen.getByLabelText(/Region/i).value).toEqual("17:1-100");
-  await userEvent.clear(screen.getByLabelText(/Region/i));
-  expect(screen.getByLabelText(/Region/i).value).toEqual("");
+  //expect(getRegionInput().value).toEqual("17:1-100");
```
adamnovak, line 28:
```suggestion
```

src/components/RegionInput.test.js
```diff
@@ -0,0 +1,44 @@
...
+  fireEvent.click(input);
+  // Key down to ensure options show up
+  fireEvent.keyDown(autocomplete, { key: "ArrowDown" });
+
+  // Make sure all our mock paths are showing up
+  expect(screen.getAllByRole("option")).toHaveLength(MOCK_PATHS.length);
+});
+test("it updates the region options", () => {
+
+});
```
adamnovak, line 44: Can you actually add this test?

adamnovak (changes requested):
The code looks pretty good, but I am not confident that the tests constrain the full setup to work. I think we need tests for:

1. Enter a region from the list and make sure the right thing is sent to `handleRegionChange`
2. Enter a region not on the list and make sure that that is actually sent to `handleRegionChange`.
3. Some kind of end-to-end test for an off-the-list region, that checks to make sure the rendered view is actually the region asked for and not any of the autocomplete options.

src/components/HeaderForm.js
```diff
@@ -165,12 +160,12 @@ class HeaderForm extends Component {
...
     } catch (error) {
       this.setState({ error: error });
-      console.log(`GET to ${this.props.apiUrl}/getFilenames failed:`, error);
+      console.error(`GET to ${this.props.apiUrl}/getFilenames failed:`, error);
     }
   };
 
   getBedRegions = async (bedFile, dataPath) => {
-    this.setState({ error: null });
+    this.setState((prevState) => ({ ...prevState, error: null }));
```
adamnovak, line 168: Why does this help at all? Doesn't `setState` implicitly preserve all the keys not mentioned?

src/components/HeaderForm.js
```diff
@@ -294,6 +276,7 @@ class HeaderForm extends Component {
             bedSelect: bedSelect,
             dataPath: dataPath,
             region: ds.region,
+            regionInfo: ds.bedFile ? this.state.regionInfo : {},
```
adamnovak, line 279: Why should this work? It looks like this is going to keep around the currently-loaded BED info if the data source we are changing to has *any* BED file set, even if it isn't the current one.

This line probably needs an explanatory comment to justify why it needs to be the way it is.

src/components/HeaderForm.js
```diff
@@ -500,41 +492,27 @@ class HeaderForm extends Component {
...
-                  <BedRegionsFormRow
-                    regionSelect={this.state.regionSelect}
-                    regionSelectOptions={this.state.regionSelectOptions}
-                    handleInputChange={this.handleInputChange}
+                {!examplesFlag && (
+                  <RegionInput
+                    pathNames={this.state.pathNames}
+                    regionInfo={this.state.regionInfo}
+                    handleRegionChange={this.handleRegionChange}
+                    region={this.state.region}
```
adamnovak, line 503: The region box appears below the dropdowns to pick files in mounted files mode, but above the file upload controls in upload mode. It should probably always appear at the bottom, and so should be added last here, just before `DataPositionFormRow`.

src/components/RegionInput.js
```diff
@@ -0,0 +1,84 @@
...
+        path + ":" + regionInfo.start[index] + "-" + regionInfo.end[index]
+      );
+    }
+    // Add descriptions
+    pathsWithRegion.push(...regionInfo["desc"]);
+  }
+
+  // Autocomplete selectable options
+  const displayRegions = [...pathsWithRegion, ...pathNamesColon];
+  //if (!region && displayRegions.length >= 1) handleRegionChange(displayRegions[0])
```
adamnovak, line 34: This shouldn't be here.

src/components/RegionInput.test.js
adamnovak: @olivewong This test still needs to be filled in to actually test what it claims to test.

adamnovak (changes requested):
This looks mostly good, but I noticed some lines that don't make sense without comments, and also one of the newly-added tests is still empty and not implemented. We should fix that before merging.

src/components/HeaderForm.js
olivewong: Reverting this back. I added it out of habit since `useState` requires copying the old properties but you're right, `setState` should preserve keys

src/components/HeaderForm.js
olivewong: The intention was that if a bedFile exists, we can keep the regionInfo was just set by getBedRegions but I'll put another call to setState if it's empty to be more clear

adamnovak: Oh OK, there are multiple `setState` calls going on here. That makes sense.

adamnovak:
I think the `"it updates the region options"` test is still missing.

Issue #198, opened 2022-11-14
Support for more graph formats


adamnovak:
Right now we can only support a .xg graph. We should be able to support other graph formats known to vg, like .vg, .pg, .hg, .gbz, and maybe even GFA files directly (if they are small enough).

`vg chunk` should already support most of these, but our import scripts, server file listing backend, and browser selection and upload front-end don't.

PR #204, opened 2022-11-23
Allow additional graph formats

ducku:
Added support for .vg, .pg, .hg, and .gbz extensions for mounted and uploaded graph files. 

- Check for files ending in extension format
- Changed xg variable names
- Changed tests to match new names
- Added new example data


src/server.js
```diff
@@ -156,6 +164,15 @@ function indexGamSorted(req, res) {
   });
 }
 
+function endsWithExtensions(file, extensions) {
```
adamnovak, line 167: It would be nice to have documentation comments for new functions and what they are supposed to accomplish.

adamnovak (approved):

adamnovak:
This fixed #198.

Issue #146, opened 2022-02-23
Fix react warnings in tests


adamnovak:
Running the tests will generate React warnings, mostly about components that are unmounted getting async calls returned and trying to redraw.

We should fix them.

PR #207, opened 2022-12-12
Fix warnings in tests

rmccrear:
Hi, vgteam. I came across your project and was intrigued, so I thought I might try to contribute a few commits. I noticed that you had an issue called **Fix react warnings in tests** #146. I've worked out the following fixes for that issue.

I'm sorry about the large number of commits. I tried to atomize each commit to deal with one warning message at a time. Please let me know if anything is unclear or if you would like this to be split up into multiple PRs.

These updates should allow the all tests to run without warnings.

6ea1bf5c4bb5f1432086bd7012f5e404d479f232 It looks like there was a typo in one of the routes. This commit simply adds a “/” character to fix the route.

99573cdca42063a1af9390cba862d42ead205895 There were some warnings about setState being called within a state update. 
In `HeaderForm.js`, the functions `getBedRegions()` and `getPathNames()` both have calls to `setState()`, so this commit moves those two method calls out from the anonymous function passed to `setState()`.

0d96db8bcfd7bf199ddaf36bd776bd610ec4a7a3 In the unit tests, fetch calls were being being made to the server and weren't being returned. This commit adds a `__mocks__` directory and mocks the `fetchAndParse()` module so that no HTTP calls are actually made during the unit tests. (But mock data *is* returned to the Components being tested) the mock function is brought in with `jest.mock("./fetchAndParse")` in `App.test.js`. (For mock data, I used the data from the default `snp1kg-BRCA1 ` except for the chunkedData, which was too large. For the chunkedData I used data from `backward.vg`, which is smaller.)

395a31bd57e7ab996bb4202699de974148dd75ce In the end-to-end tests, there were warnings about components that are unmounted trying to update. It turns out that there were calls to fetch that were returning after the components had already unmounted. To fix this warning, fetch requests can be canceled before they complete. We can cancel them and then avoid updating state in componentWillUnmount. (Note: There may be other ways to deal with this issue if you don't want to add the complexity of canceling fetch calls.)

321479eb114e0605ad95c42d8f531bfa0c746e8f In `server.js` there were some `==` operators that should be changed to `===` and other minor typos.

e708a2120b9119c581d63ef338a48fad2851fd05 This puts in some additional tests to check error states when the client receives errors from the server. The tests led to a few small fixes regarding error handling in TubeMapContainer. (Note, that this will also cause the console to log those mock errors during testing.)

312cdc887af90da960680ad8672838c3532aab6c The tests were giving an error about `inline` being used on `Form`. ReactStrap no longer supports `inline-form`. So this removes the inline prop from Forms and updates the CSS rule to put margins on buttons. (See here for details: 
https://github.com/reactstrap/reactstrap/commit/5359bc5982c4d0c2440666945b9f9d52f0961200)

In the end-to-end tests, the server hangs when the tests complete, and the test runner complained about that. We can wait on the HTTP server to exit by waiting on its callback function. f6877e6d6e31956538a219a43a13bc5eba0984c6 Unfortunately, waiting on the server to finish afterEach test makes the tests run considerably slower, so one solution is to start the server once beforeAll tests, instead of beforeEach test. fa2c7da2d98e0d59775abf93bd8bb75a77879866

37f3e6f8219be313ef2d8acc3d777df81a8af1fe There is also a depreciation warning about using `Buffer()`. This can be switched to `Buffer.alloc()`. See these links for details:
https://nodejs.org/en/docs/guides/buffer-constructor-deprecation/
https://stackoverflow.com/questions/52165333/deprecationwarning-buffer-is-deprecated-due-to-security-and-usability-issues


adamnovak:
Thanks for figuring this all out @rmccrear! This all looks great!

I'm not sure we want *quite* so much mock data for the no-server tests; we ought to be able to get by with more or less empty replies there. But we can cut them down later if we want.

rmccrear:
There's probably a way to reduce the amount of mock data. I just used the smallest data sets that I could find already in the project. But, a minimal test example would definitely be better.

PR #208, opened 2022-12-14
GBZ Support

ducku:
Added basic track structure and GBZ support

- Track structure containing Array<file>
- Comparison of filenames on Go button press (see if rerender is needed)
- Allow for gbz files as both graph and haplotype

GBZ file support only works when the file contains generic or reference paths from a base graph. 
[Building a Useful GBZ](https://github.com/vgteam/vg/wiki/VG-GBWT-Subcommand#building-a-useful-gbz)


npm-debug.log
```diff
@@ -0,0 +1,40 @@
+0 info it worked if it ends with ok
```
adamnovak, line 1: You probably don't want to commit this file.

src/components/HeaderForm.js
```diff
@@ -38,6 +45,7 @@ const CLEAR_STATE = {
 
   pathNames: ["none"],
 
+  track: undefined,
```
adamnovak, line 48:
```suggestion
  tracks: [],
```

src/components/HeaderForm.js
```diff
@@ -68,6 +76,40 @@ const EMPTY_STATE = {
...
+  //remove empty files here?
+
+  const track = {
+    files: files,
+  };
+  return track;
+}
+
+// Checks if all file names in the track are equal
+function tracks_equal(curr, next) {
```
adamnovak, line 95: Do we not care if the types of the tracks differ?

Could we maybe use `isEqual` from `react-fast-compare` which we already depend on? Or do we need our own function so we can later ignore settings changes on the tracks that won't affect the downloaded data?

src/components/HeaderForm.js
```diff
@@ -300,23 +342,35 @@ class HeaderForm extends Component {
...
     if (this.props.dataOrigin !== dataOriginTypes.API) {
       this.props.setColorSetting("haplotypeColors", "ygreys");
       this.props.setColorSetting("forwardReadColors", "reds");
     }
-    const viewTarget = this.getNextViewTarget();
-    this.props.setCurrentViewTarget(viewTarget);
+
+    for (let i = 0; i < this.props.getCurrentViewTarget().tracks.length; i++) {
+      if (!tracks_equal(this.props.getCurrentViewTarget().tracks[i], this.getNextViewTarget().tracks[i])){
+        const viewTarget = this.getNextViewTarget();
```
adamnovak, line 364: I don't think we want to call `getNextViewTarget()` so many times when we don't need to; it does some actual work if I remember right. `getCurrentViewTarget()` does less but still seems like work.
```suggestion

    const viewTarget = this.getNextViewTarget();
    const oldViewTarget = this.props.getCurrentViewTarget();
    for (let i = 0; i < oldViewTarget.tracks.length; i++) {
      if (!tracks_equal(oldViewTarget.tracks[i], viewTarget.tracks[i])){
```

adamnovak: *Eventually* we're also going to need to handle the case of adding a track.

src/components/HeaderForm.js
```diff
@@ -300,23 +342,35 @@ class HeaderForm extends Component {
     }
   };
   getNextViewTarget = () => ({
+    tracks: new Array(create_track({name: this.state.graphFile, type: fileTypes.GRAPH}), 
+                      create_track({name: this.state.gbwtFile, type: fileTypes.HAPLOTYPE}), 
+                      create_track({name: this.state.gamFile, type: fileTypes.READ}), 
+                      create_track({name: this.state.bedFile, type: fileTypes.BED})),
```
adamnovak, line 348: Shouldn't we only include the tracks we are actually using?

Also, I'm not sure that the BED files are *really* being used as tracks here; I think they are more like caches, and not BED files we display as regions on the visualization.

src/end-to-end.test.js
```diff
@@ -256,7 +256,7 @@ describe("When we wait for it to load", () => {
 it("produces correct link for view before & after go is pressed", async () => {
   // First test that after pressing go, the link reflects the dat form
   const expectedLinkBRCA1 =
-    "http://localhost?name=snp1kg-BRCA1&graphFile=snp1kg-BRCA1.vg.xg&gamFile=NA12878-BRCA1.sorted.gam&dataPath=default&region=17%3A1-100&bedFile=snp1kg-BRCA1.bed&dataType=built-in";
+    "http://localhost?tracks=%5Bobject+Object%5D%2C%5Bobject+Object%5D%2C%5Bobject+Object%5D%2C%5Bobject+Object%5D&name=snp1kg-BRCA1&graphFile=snp1kg-BRCA1.vg.xg&gamFile=NA12878-BRCA1.sorted.gam&dataPath=default&region=17%3A1-100&bedFile=snp1kg-BRCA1.bed&dataType=built-in";
```
adamnovak, line 259: This can't actually work; the old link contained the information that we should use e.g. `snp1kg-BRCA1.vg.xg`, while the new link only says `[object Object]`. We might be missing some proper serialization and deserialization of the track objects' fields.

src/server.js
```diff
@@ -193,27 +214,31 @@ api.post("/getChunkedData", (req, res, next) => {
   // This request owns the directory, so clean it up when the request finishes.
   req.rmChunk = true;
 
-  // We always have an graph file
-  const graphFile = req.body.graphFile;
+  //TODO: loop through and process all tracks
```
adamnovak, line 217: I guess it is going to pass tests to rely on an exact fixed order of tracks, but I don't think we really want to do it that way.

src/server.js
```diff
@@ -305,7 +330,55 @@ api.post("/getChunkedData", (req, res, next) => {
...
+    if (graphFile.endsWith(".gbz")) {
+      if (req.withGbwt) {
+        if (gbwtFile.endsWith(".gbz")) {
+          vgChunkParams.push("-x", `${dataPath}${graphFile}`);
+        } else {
+          vgChunkParams.push("-x", "--ignore-embedded-haplotypes", `${dataPath}${graphFile}`);
+        }
+      }
+    }
+*/
```
adamnovak, line 347: This code is commented out and so ought to be removed.

src/server.js
```diff
@@ -305,7 +330,55 @@ api.post("/getChunkedData", (req, res, next) => {
...
+    if (req.withGbwt) {
+      //either push gbz with graph and haplotype or push seperate graph and gbwt file
+      if (graphFile.endsWith(".gbz") && gbwtFile.endsWith(".gbz")) { 
+        // use gbz haplotype
+        vgChunkParams.push("-x", `${dataPath}${graphFile}`);
```
adamnovak, line 352: Should we maybe check to make sure these are really the same file?

src/server.js
```diff
@@ -331,6 +405,7 @@ api.post("/getChunkedData", (req, res, next) => {
       // Use a GBWT haplotype database
       vgChunkParams.push("--gbwt-name", `${dataPath}${gbwtFile}`);
     }
+    */
```
adamnovak, line 408: We should cut this code as well if it's not needed.

src/Types.ts
```diff
@@ -9,10 +9,27 @@ type DataPath = "mounted" | "default" | "upload";
 // Fills input for the "Data:" dropdown in the HeaderForm
 type DataType = "built-in" | "file-upload" | "mounted files" | "examples";
 
+// Possible filestypes taken from the request
+// Files like GBZ contains graph and maybe haplotype
+type filetype = "graph" | "haplotype" | "read" | "bed" | "graph-haplotype";
```
adamnovak, line 14: If we're using the GBZ file with graph type and with haplotype type to indicate whether we want its haplotypes or not, we don't need the combined type anymore.
```suggestion
// Files like GBZ contains graph and maybe haplotype and so can be either
type filetype = "graph" | "haplotype" | "read" | "bed";
```

adamnovak (changes requested):
This looks pretty good!

It looks like it breaks link-to-view, because we get `[object Object]` in the link instead of any information about the contents of the track objects. I'm not sure if the tests are clever enough to catch that.

It also seems to be hardcoded to 4 tracks in the given order, which will *work* for now, but might be worth fixing before it is quite needed.

src/Types.ts
```diff
@@ -9,10 +9,27 @@ type DataPath = "mounted" | "default" | "upload";
...
+}
+
 // Describes something the Tube Map can look at, specifically a region and the files the region is in.
 // These parameters can be extracted from defaults (config), Header Form input, or URL parameters
 type ViewTarget = {
   region: string; // Format: path:start-end
+  tracks: Array<track>;
   graphFile: string;
   gbwtFile?: string;
   gamFile?: string;
```
adamnovak, line 35: Isn't the array replacing some of these?

src/components/CopyLink.js
```diff
@@ -22,8 +24,11 @@ export function CopyLink(props) {
     // Turn viewTarget into a URL query string
     const viewTarget = props.getCurrentViewTarget();
 
-    const params = new URLSearchParams(viewTarget).toString();
+    console.log("copying");
+    //const params = new URLSearchParams(viewTarget).toString();
```
adamnovak, line 28:
```suggestion
```

adamnovak, line 28:
```suggestion
```

src/components/CopyLink.js
```diff
@@ -22,8 +24,11 @@ export function CopyLink(props) {
...
+    const params = qs.stringify(viewTarget, { encode: false });
```
adamnovak, line 29: Is there a reason we need to turn off `encode` here? That ought to be explained in a comment.

src/components/CopyLink.js
```diff
@@ -46,18 +51,36 @@ export const urlParamsToViewTarget = (url) => {
...
+        // If multiple values under the same key is recieved
+        if (Object.prototype.hasOwnProperty.call(result, key)) {
+          if (typeof result[key] === 'string') {
+            result[key] = [result[key]];
+          }
+          result[key].push(value);
+        } else {
+          result[key] = value;
+        }
+      }
```
adamnovak, line 81: I think that, when we split the URL on `?` and we don't have anything after a first `?`, then the URL's `searchParams` are also always empty, because if we had them they would be after `?`. So I think we would always hit the `return null` in the `else` and never hit most of the logic in the `else`.

Also, we don't need this whole block, because `qs.parse()` just implements all this stuff for us, right?

We could probably simplify this down to:

```suggestion
  const params = new URL(url).searchParams;
  if (params) {
    result = qs.parse(params);
  }
```

Is there a reason why that wouldn't work?

src/config.json
```diff
@@ -2,6 +2,10 @@
   "BACKEND_URL": false,
   "DATA_SOURCES": [
     {
+      "tracks": [{"files": [{"name": "snp1kg-BRCA1.vg.xg", "type": "graph"}]},
+                 {"files": [{"name": null, "type": "haplotype"}]},
+                 {"files": [{"name": "NA12878-BRCA1.sorted.gam", "type": "read"}]}],
+                   
       "name": "snp1kg-BRCA1",
       "graphFile": "snp1kg-BRCA1.vg.xg",
```
adamnovak, line 10: The original `xxxFile` fields (except `bedFile`) aren't used anymore, right? We should probably drop them.

src/end-to-end.test.js
```diff
@@ -256,7 +256,7 @@ describe("When we wait for it to load", () => {
 it("produces correct link for view before & after go is pressed", async () => {
   // First test that after pressing go, the link reflects the dat form
   const expectedLinkBRCA1 =
-    "http://localhost?name=snp1kg-BRCA1&graphFile=snp1kg-BRCA1.vg.xg&gamFile=NA12878-BRCA1.sorted.gam&dataPath=default&region=17%3A1-100&bedFile=snp1kg-BRCA1.bed&dataType=built-in";
+    "http://localhost?tracks[0][files][0][name]=snp1kg-BRCA1.vg.xg&tracks[0][files][0][type]=graph&tracks[1][files][0][name]=&tracks[1][files][0][type]=haplotype&tracks[2][files][0][name]=NA12878-BRCA1.sorted.gam&tracks[2][files][0][type]=read&name=snp1kg-BRCA1&graphFile=snp1kg-BRCA1.vg.xg&gamFile=NA12878-BRCA1.sorted.gam&dataPath=default&region=17:1-100&bedFile=snp1kg-BRCA1.bed&dataType=built-in";
```
adamnovak, line 259: I guess we need to turn off encoding so we get the real `:` here and the resulting URL is easier to read?

src/server.js
```diff
@@ -173,6 +185,15 @@ function endsWithExtensions(file, extensions) {
   return false;
 }
 
+function getFileFromType(track, type) {
```
adamnovak, line 188: This function should probably have a doc comment explaining that e.g. if not found you get "none" and not `null` or something falsey.

src/server.js
```diff
@@ -305,7 +330,45 @@ api.post("/getChunkedData", (req, res, next) => {
...
+      // push graph file
+      if (graphFile.endsWith(".gbz")) {
+        vgChunkParams.push("-x", `${dataPath}${graphFile}`, "--no-embedded-haplotypes");
+      } else {
+        vgChunkParams.push("-x", `${dataPath}${graphFile}`);
+      }
+    }
+    
+    
+    //vgChunkParams.push("-x", `${dataPath}${graphFile}`);
```
adamnovak, line 371:
```suggestion
```

adamnovak (changes requested):
This is improved! But I found some dead code that I think we should cut before merging, and a few places where I think more comments are needed.

Also I think most of the `whateverFile` fields in the view target are now unused and should be cut.

src/end-to-end.test.js
ducku: Right, I'm forgetting the exact URL format with encoding enabled, but I believe I found this format of URL to be more fitting.

ducku:
The ViewTarget object no longer uses the xxxFile format, but the API still uses the xxxFile format as that's what the UI is representing. Definitely still see some "to be dead" code that I'm hoping to replace as we change the UI.

adamnovak (approved):

PR #213, opened 2023-01-23
Allow 2 Read Files to be Inputted

ducku:
- Mounted and file upload page can input additional GAM file
- Reads are processed and concatenated
- Added cactus example data with first 20 reads split between 2 files
     - End-to-end test for verifying example


adamnovak:
@ducku I merged this in with #220.

PR #220, opened 2023-01-30
Jump straight to Node 18

adamnovak:


adamnovak:
This should bring CI up to Node 18, which is the current LTS major version, and should temporarily stop us from having to deal with all the other modules dropping our Node (until it becomes too old again...).

When we go to deploy the online demo again, we're going to have to set up the demo server with NVM to get the right Node version.

Issue #222, opened 2023-01-31
Implement `TrackFilePicker`


adamnovak:
We need a `TrackFilePicker` as defined in #205. Fir a given track type it will let you choose between available files that can make that type of track.

PR #232, opened 2023-03-07
Track file picker

ducku:
File dropdown selection component alongside tests and demo

- would like to get redux to work 
- need to change file submission endpoints to get file upload working

### DIFF
--- a/src/components/TrackFilePicker.demo.js
+++ b/src/components/TrackFilePicker.demo.js
@@ -7,16 +7,21 @@ import TrackFilePicker from "./TrackFilePicker";
 
 export default (<Demo
   props={{
-    fileOptions: P.json(["fileA", "fileB", "fileC", "anotherfile.vg"]),
-    fileSelect: P.string("fileA"),
-    pickerType: P.string("dropdown"),
+    tracks: P.json([{"files": [{"name": "fileA1.vg", "type": "graph"},
+                              {"name": "fileA2.gbwt", "type": "haplotype"}]},
+                    {"files": [{"name": "fileB1.gbwt", "type": "haplotype"},
+                              {"name": "fileB2.gam", "type": "read"}]},
+                    {"files": [{"name": "fileC1.xg", "type": "graph"}]}]),
+
+    fileType: P.choices(["graph", "haplotype", "read"]),
+    pickerType: P.choices(["dropdown", "upload"]),
   }}
 >
   {
     (props, update) => {
       return <TrackFilePicker {...props} handleInputChange={(file) => {
         // takes file name as value, updates the fileSelect prop in TrackFilePicker
-        update({fileSelect: file});
+        update({value: file});
       }}/>
     }
   }
--- a/src/components/TrackFilePicker.demo.js
+++ b/src/components/TrackFilePicker.demo.js
@@ -13,6 +13,7 @@ export default (<Demo
                               {"name": "fileB2.gam", "type": "read"}]},
                     {"files": [{"name": "fileC1.xg", "type": "graph"}]}]),
 
+    value: P.json({"name": "fileA1.vg", "type": "graph"}),
     fileType: P.choices(["graph", "haplotype", "read"]),
     pickerType: P.choices(["dropdown", "upload"]),
   }}
--- a/src/components/TrackFilePicker.demo.js
+++ b/src/components/TrackFilePicker.demo.js
@@ -1,21 +1,25 @@
 import Demo, {props as P} from 'react-demo'
-import React from "react";
 // See https://github.com/rpominov/react-demo for how to make a demo
+
+import React from "react";
 
 import TrackFilePicker from "./TrackFilePicker";
 
 export default (<Demo
   props={{
-    fileOptions: P.json(["fileA", "fileB", "fileC"]),
+    fileOptions: P.json(["fileA", "fileB", "fileC", "anotherfile.vg"]),
     fileSelect: P.string("fileA"),
+    pickerType: P.string("dropdown"),
   }}
 >
   {
     (props, update) => {
-      // We need to render the component under test using the props in props, and
-      // call update when the component wants to adjust the props.
-      return <TrackFilePicker {...props}/>
+      return <TrackFilePicker {...props} handleInputChange={(file) => {
+        // takes file name as value, updates the fileSelect prop in TrackFilePicker
+        update({fileSelect: file});
+      }}/>
     }
   }
 </Demo>)
+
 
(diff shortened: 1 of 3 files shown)


src/components/TrackFilePicker.demo.js
```diff
@@ -0,0 +1,25 @@
...
+
+import React from "react";
+
+import TrackFilePicker from "./TrackFilePicker";
+
+export default (<Demo
+  props={{
+    fileOptions: P.json(["fileA", "fileB", "fileC", "anotherfile.vg"]),
+    fileSelect: P.string("fileA"),
+    pickerType: P.string("dropdown"),
```
adamnovak, line 12:
```suggestion
    pickerType: P.choices(["dropdown", "upload"]),
```

This will let the type value not have to go through unacceptable intermediate states like "dropdo" as the user is typing.

src/components/TrackFilePicker.demo.js
```diff
@@ -0,0 +1,25 @@
+import Demo, {props as P} from 'react-demo'
+// See https://github.com/rpominov/react-demo for how to make a demo
+
+import React from "react";
+
+import TrackFilePicker from "./TrackFilePicker";
+
+export default (<Demo
+  props={{
+    fileOptions: P.json(["fileA", "fileB", "fileC", "anotherfile.vg"]),
```
adamnovak, line 10: I think instead of just the filenames, we wanted the input to be a collection of available tracks (so objects including a file and also a track type), right?

src/components/TrackFilePicker.js
```diff
@@ -0,0 +1,76 @@
...
+    fileOptions: PropTypes.array.isRequired,
+    fileSelect: PropTypes.string.isRequired,
+    handleInputChange: PropTypes.func.isRequired,
+    pickerType: PropTypes.string,
+    className: PropTypes.string,
+}
+  
+TrackFilePicker.defaultProps = {
+  pickerType: "dropdown",
+  classname: undefined
```
adamnovak, line 72:
```suggestion
  className: undefined
```

The capitalization has to match right?

src/components/TrackFilePicker.js
```diff
@@ -0,0 +1,76 @@
...
+import PropTypes from "prop-types";
+import Select from "react-select";
+
+
+
+//handleInputChange takes in a string trackType
+export const TrackFilePicker = ({
+  fileOptions, // array of string representing files names
+  fileSelect, 
+  handleInputChange,
```
adamnovak, line 11: Should we call this `onChange` to match the naming in the `TrackTypeDropown`?

Also, do we want it to take a whole track object instead of just a filename?

adamnovak (changes requested):
This looks pretty good, except I thought we wanted to put the responsibility for filtering the available tracks onto this control. It doesn't currently have an input prop to take a track type like "read" or "graph", and it can't throw out input files from `fileOptions` that aren't the right kind of track.

@ducku Do you think it is better to do it like this, where this control just picks a filename from a set of filenames given to it already, and then push the responsibility for filtering down the list of filenames and for composing a track object into the parent `TrackListItem` component?

It might work better that way because it would be easier to bind the type of the track down to the type picker and the filename of the track down to the file name picker, and the settings of the track down to the settings button. But I'm not sure if you are doing it this way because you have an idea for how this is going to fit together like that and think it is better, or because the issue in #222 wasn't clear that the filtering was envisioned to be inside here.

src/components/TrackFilePicker.test.js
```diff
@@ -0,0 +1,88 @@
...
+        fireEvent.keyDown(fileSelectComponent.firstChild, {key: "ArrowDown"});
+        await waitFor(() => getByText("fileA2.gbwt"));
+        fireEvent.click(getByText("fileA2.gbwt"));
+
+        expect(fakeOnChange).toHaveBeenCalledTimes(2);
+        expect(fakeOnChange).toHaveBeenCalledWith({"name": "fileA2.gbwt", "type": "haplotype"});
+
+    });
+
+    it('should call onChange when queried by input value' ,async() => {
```
adamnovak, line 63: I don't think we are querying by input value here. We are querying just for the `input` I think by tag name using a selector, and then we are *firing an event* to set the value.

```suggestion
    it('should call onChange when the input value is changed' ,async() => {
```

src/components/TrackFilePicker.js
```diff
@@ -0,0 +1,79 @@
+import PropTypes from "prop-types";
+import Select from "react-select";
+
+
+
+//handleInputChange takes in a string trackType
+export const TrackFilePicker = ({
+  tracks, // array of available tracks
```
adamnovak, line 8: What format should this array be in? Can you refer back to the types file? Because the demo example value doesn't look that much like a real set of available tracks.

src/components/TrackFilePicker.demo.js
```diff
@@ -0,0 +1,30 @@
...
+
+import TrackFilePicker from "./TrackFilePicker";
+
+export default (<Demo
+  props={{
+    tracks: P.json([{"files": [{"name": "fileA1.vg", "type": "graph"},
+                              {"name": "fileA2.gbwt", "type": "haplotype"}]},
+                    {"files": [{"name": "fileB1.gbwt", "type": "haplotype"},
+                              {"name": "fileB2.gam", "type": "read"}]},
+                    {"files": [{"name": "fileC1.xg", "type": "graph"}]}]),
```
adamnovak, line 14: Multi-file tracks like this are good to test, but they might be a bit counterintuitive, since despite the "TrackFilePicker" name I think the right behavior here would be to allow selecting any whole track object with a file of a matching type.

adamnovak (changes requested):
I like the track filtering much better, but now we only have one-way, upwards binding: adjusting the control in the pace will make a call upwards with the newly selected track, but there's no way to tell the TrackFilePicker what ought to be selected in it when e.g. setting it up.

If we want to load up a preset example collection of tracks, maybe when opening a link to a view, we need to be able to set the set of currently selected tracks and have the track selection UI render with those particular tracks displayed as selected. So the `TrackFilePicker` needs to have a prop for the currently selected track, in addition to calling `handleInputChange` or wehatever the handler will be called when the user changes the selection.

It looks like right now we are meant to take as the track set a collection of track objects, but the selected value is reported just a single filename, and if there are multi-file tracks available the files are treated as separately selectable. I think what we *really* want is to take and report the currently selected value as a whole track object, with files array inside it. It's fine if we only support single-file tracks to start, but eventually we'd probably want to handle multi-file tracks: come up with something to dsiplay them as in the dropdown and work out how their multiple types interact with filtering.

adamnovak (approved):
I pulled this down to test and it still wasn't *quite* right; changing the value in the demo to a different file didn't actually change the currently selected file on the right side in the actual dropdown.

I added a test for that rerender behavior, and I fixed it by changing from `defaultValue` to `value` as the input to React-Select, and by providing it a `getOptionValue` function it can use to actually do comparisons when the `value` is specified as a distinct object that is merely deep-equal to the one we mean in the `options` list.

Now I think this is ready.

PR #244, opened 2023-03-20
adding track delete button files

shreyasun:
Work in progress for track delete button


adamnovak (approved):
I patched up the test to do a click instead of trying to do a dropdown select, and I think this works now.

PR #245, opened 2023-03-27
Track settings button

ducku:
- Track settings and wrapper button 
- Tests and demo

Not too sure if the settings and button component should have separate tests and demos. Might be better to merge their demo and tests?


adamnovak (approved):
This looks pretty good!

It might indeed make sense to merge the demos, but I don't think the duplication is going to be too big of a problem for now.

I just dropped the executable bit being set on some of the files, but other than that I think we can merge this.

I also noticed that, when adding the multiple GAM files, we ended up with controls for reverse-strand haplotype paths, which I don;t think do anything because haplotypes don't really have a notion of being reverse-strand. So I might try and pull those out, and then get rid of the "Forward" language here.

PR #246, opened 2023-03-28
Plug in track settings

adamnovak:
This plugs in the track settings forms and gets rid of the reverse haplotype color setting.

I had to change the track settings dialogs to modals because of https://github.com/yjose/reactjs-popup/issues/334, and they don't look as good like that, because then the position is ignored and they just end up in the center of the screen.


adamnovak:
I moved the card stuff to the popup and added a label string for the track settings so they could be used inline like this:
image

Issue #248, opened 2023-04-04
TrackSettingsButton should look nicer


adamnovak:
It should look like `TrackDeleteButton`

PR #249, opened 2023-04-04
Change appearance of track settings button and make popup controlled

shreyasun:
For issue #248

### DIFF
--- a/src/components/TrackSettingsButton.js
+++ b/src/components/TrackSettingsButton.js
@@ -1,7 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import Popup from 'reactjs-popup';
 import PropTypes from "prop-types";
 import TrackSettings from "./TrackSettings.js";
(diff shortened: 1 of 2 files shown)


src/components/TrackSettingsButton.js
```diff
@@ -15,8 +19,17 @@ export const TrackSettingsButton = ({
     label,
     availableColors
 }) => {
+    // based off of https://react-popup.elazizi.com/controlled-popup/#using-open-prop
+    const [open, setOpen] = useState(false);
+    const close = () => setOpen(false);
     return(
-        <Popup trigger={<button type="button"> Settings </button>} position="right center" contentStyle={{width: "760px"}} modal>
+      <div>
```
adamnovak, line 26: I think this `<div>` and `</div>` could be replaced with `<>` and `</>` and save an unnecessary HTML element, but it should work fine as is.

adamnovak (approved):

PR #251, opened 2023-04-09
Track list item

ducku:
TrackListItem component combining:
- TrackFilePicker
- TrackTypeDropdown
- TrackDeleteButton
- TrackSettingsButton

Changing the track type clears the file field, preventing invalid track type, file type combinations.

Styling may still need a bit of work.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -191,9 +191,14 @@ input[type="file"] {
   }
 }
 
-.class-col {
+.tracklist-button {
   display: flex;
-  align-items: center;
-  justify-content: center;
+  align-items: left;
+  justify-content: left;
   height: 100%;
+  margin-left: 10px;
+}
+
+.tracklist-dropdown {
+  margin-top: 4px;
 }
--- a/src/App.css
+++ b/src/App.css
@@ -190,3 +190,10 @@ input[type="file"] {
     opacity: 1;
   }
 }
+
+.class-col {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+}
--- a/src/components/TrackListItem.demo.js
+++ b/src/components/TrackListItem.demo.js
@@ -7,23 +7,28 @@ import TrackListItem from "./TrackListItem";
 
 export default (<Demo
   props={{
-    trackType: P.choices(["graph", "haplotype", "read"]),
-    availableTracks: P.json([{"files": [{"name": "fileA1.vg", "type": "graph"},
-                                        {"name": "fileA2.gbwt", "type": "haplotype"}]},
-                             {"files": [{"name": "fileB1.gbwt", "type": "haplotype"},
-                                        {"name": "fileB2.gam", "type": "read"}]},
-                             {"files": [{"name": "fileC1.xg", "type": "graph"}]}]),
+    trackProps: P.json({
+      trackFile: undefined,
+      trackType: "graph",
+      availableTracks: [{"files": [{"name": "fileA1.vg", "type": "graph"},
+                                          {"name": "fileA2.gbwt", "type": "haplotype"}]},
+                               {"files": [{"name": "fileB1.gbwt", "type": "haplotype"},
+                                          {"name": "fileB2.gam", "type": "read"}]},
+                               {"files": [{"name": "fileC1.xg", "type": "graph"}]}],
+      trackColorSettings: {    
+        mainPallete: "blues",
+        auxPallete: "reds",
+        colorReadsByMappingQuality: false},
+    })
   }}
 >
   {
     (props, update) => {
       return <TrackListItem {...props} 
-        onChange={(newTrackType, newTrackFile, newTrackColorSettings) => {
-          // Bind new value back up to value.
-          // Remember: we get a string 
-          update({trackType: newTrackType});
-          update({trackFile: newTrackFile});
-          update({trackColorSettings: newTrackColorSettings})}} 
+        onChange={(newTrackProps) => {
+
+          update({trackProps: newTrackProps});
+        }} 
         onDelete={() => {console.log("track deleted")}}/>
     }
   }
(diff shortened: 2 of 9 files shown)


src/components/TrackListItem.js
```diff
@@ -0,0 +1,94 @@
...
+    Row,
+    Col
+  } from "reactstrap";
+import {TrackFilePicker} from './TrackFilePicker';
+import {TrackTypeDropdown} from './TrackTypeDropdown';
+import {TrackDeleteButton} from './TrackDeleteButton';
+import {TrackSettingsButton} from './TrackSettingsButton';
+
+
+export const TrackListItem = ({
```
adamnovak, line 13: There needs to be some documentation here for what `onChange` has to be written to expect. Otherwise it will be hard to figure out what you need to do to use the component.

Right now it looks like onChange needs to accept 3 arguments: a string track type, a string track filename, and an object of the color settings type, all in that order. It also looks like right now it needs to be able to cope with `undefined` values, but I want it to be protected from ever seeing them instead.

I guess if the track type, file, and color settings all come in as separate props, that makes sense.

If we ever switch to just one prop for the track object as a whole, we'd then want to switch to a function that takes a modified copy of the track object.

src/App.css
```diff
@@ -190,3 +190,10 @@ input[type="file"] {
     opacity: 1;
   }
 }
+
+.class-col {
```
adamnovak, line 194: This should probably have a more specific name; we don't want all the CSS class names to start with "class", and it applies to only a couple of specific "Col"s in the TrackListItem.

src/components/TrackListItem.demo.js
```diff
@@ -0,0 +1,32 @@
...
+import TrackListItem from "./TrackListItem";
+
+export default (<Demo
+  props={{
+    trackType: P.choices(["graph", "haplotype", "read"]),
+    availableTracks: P.json([{"files": [{"name": "fileA1.vg", "type": "graph"},
+                                        {"name": "fileA2.gbwt", "type": "haplotype"}]},
+                             {"files": [{"name": "fileB1.gbwt", "type": "haplotype"},
+                                        {"name": "fileB2.gam", "type": "read"}]},
+                             {"files": [{"name": "fileC1.xg", "type": "graph"}]}]),
```
adamnovak, line 15: Shouldn't we start with a `trackFile` and `trackColorSettings` here as well?

adamnovak (changes requested):
This looks pretty good, but there's some stuff that should be renamed and commented.

I also think that `onChange` for the `TrackListItem` should be protected from ever being called with missing values. The component itself should deal with having to keep around partially-completed or mutually-inconsistent settings of its child components, and maybe turn red or something, and then only send them up the component tree when they make sense together.

src/components/TrackTypeDropdown.js
```diff
@@ -14,7 +16,7 @@ export function TrackTypeDropdown (props) {
     // eventToString: given a change event on a dropdown with various options, return the 
     //string value for the user's selection 
     const eventToString = (changeEvent) => {
-      return changeEvent.target.value;
+      return changeEvent.value;
```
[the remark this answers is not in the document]
ducku: Was in a bit of a rush to change the component to use React Select for it to look consistent. You're right, the functions don't really make sense anymore.

src/components/TrackListItem.demo.js
ducku: I struggled a bit with the design of default variables, and whether/when to use them. I wasn't sure which components should have them and what props should have a default. I omitted them here out of readability I suppose.

src/components/TrackListItem.js
```diff
@@ -0,0 +1,119 @@
...
+import {TrackDeleteButton} from './TrackDeleteButton';
+import {TrackSettingsButton} from './TrackSettingsButton';
+import React, { useEffect, useReducer } from 'react';
+
+
+export const TrackListItem = ({
+    trackProps,
+    // trackProps expects an object with
+      // trackType: string
+      // trackFile: file object
```
adamnovak, line 18: Is this allowed to be undefined, since it is in the demo?

src/components/TrackListItem.js
```diff
@@ -0,0 +1,112 @@
...
+      newTrackColorSettings[key] = value;
+      dispatch({type: "setColorSettings", payload: newTrackColorSettings});
+    }
+
+    // https://stackoverflow.com/questions/55840294/how-to-fix-missing-dependency-warning-when-using-useeffect-react-hook
+    // useEffect hook to tell react to call onchange after state changes
+    useEffect(() => {
+      if (myTrackProps.trackFile !== undefined){
+        _onChange.current(myTrackProps);
+      }
```
adamnovak, line 68: Isn't this going to call onChange on the first render, as long as `trackFile` is set in the props?

(And also if the user changes the track type, then changes it back and re-picks the initial file? Though that might be OK).

I feel like we should only call onChange when there's an actual change. So maybe we want to skip the call if myTrackProps is deep-equal to trackProps?

adamnovak (changes requested):

src/components/TrackListItem.js
```diff
@@ -0,0 +1,113 @@
...
+
+
+export const TrackListItem = ({
+    trackProps,
+    // trackProps expects an object with
+      // trackType: string
+      // trackFile: file object / undefined
+      // availableTracks: array of tracks(see types.ts)
+      // trackColorSettings: object(aka. colorScheme)
+      // availableColors: array of ColorPalletes
```
adamnovak, line 21: I think we don't want to have the available options as part of the `trackProps` object; we probably want everything in `trackProps` to be about the track that this controls. Otherwise we're going to end up having to answer questions like, "What if `onChange` is called and it is trying to change the list of available colors?".

adamnovak (approved):
I think we're going to need to move `availableColors` and `availableTracks` out of `trackProps` so we don't have to deal with it being syntactically possible for the control to try to change them (and since semantically they aren't properties of any track). But that should be pretty easy, and I think the update logic looks right now.

PR #257, opened 2023-04-25
Replaced preexisting help button with new help button component that …

### LINKED ISSUE #255: Use new help component to replace `?` tooltip button with a dialog
There should be a new Markdown file with at least the current tooltip text, which pops up when you click the new button, which should take the place of the old button.

We can probably get rid of the old tooltip-producing component, since it won't be needed anymore.

shreyasun:
Fixes #255

### DIFF
--- a/public/help/help-tooltip.md
+++ b/public/help/help-tooltip.md
@@ -1,0 +1,1 @@
+Seach for a coordinate range (e.g. "chr1:1-100"), a node ID ranges (e.g. "node:100-110"), a start position and a distance (e.g. "chr1:1+100"), or a node ID anchor and a distance (e.g. "node:100+10").
--- a/src/components/DataPositionFormRow.js
+++ b/src/components/DataPositionFormRow.js
@@ -8,8 +8,8 @@ import {
   faStepForward,
   faSearchPlus,
   faSearchMinus,
-  faQuestionCircle,
 } from "@fortawesome/free-solid-svg-icons";
+import HelpButton from "./HelpButton.js"
 import * as tubeMap from "../util/tubemap";
 
 const ZOOM_FACTOR = 2.0;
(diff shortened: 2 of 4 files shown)


adamnovak (approved):

rmccrear:
All the tests in "● When we wait for it to load" failed.

This might actually be just a flaky test, as I wasn't able to reproduce it locally or on my own fork on GitHub Actions. (It's possible it just timed out before the server started in the container.) 
You might try hitting "Re-run all jobs" on the [Action page](https://github.com/vgteam/sequenceTubeMap/actions/runs/4801290717) a few times to check whether or not this is the case.

By the way, would there be interest in adding a compact test reporter to allow for summarizing of the testing output? If so, I could install a test reporter and submit a PR for that. It might make the output easier to parse.

adamnovak:
@rmccrear It would be great to have a better test reporter! I feel like the current setup where we get about 5 lines for every log message during every test is suboptimal, and the default reports rely a lot on color to help you find the start of each test log or the status of each test, and that doesn't render on Github Actions.

adamnovak:
I set this to rerun to make sure this is repeatable.

rmccrear:
I was mistaken about the flaky tests. You are correct.
Adding a catch to the promise chain would solve the problem.
Something like: 

      // src/components/HelpButton.js

      fetch(file)
        .then(res => res.text())
        .then(md => { setContent(md) })
        .catch((e) => {
          setContent('Fallback Help Text');
        });
    }, [file]);

BTW, I've opened PR https://github.com/vgteam/sequenceTubeMap/pull/261 with the test reporter. (It ends up showing up in color on GitHub, as well.)

rmccrear:
What is happening is that the end-to-end tests do not start up the whole frontend webserver, but they do start up the backend server. Since the `.md` files are served by the frontend, when you fetch them, it fails because the full frontend is not available.

So, the end-to-end tests can do a "real" fetch to the backend, but can't do a fetch to access a resource like an `.md` file served by the frontend. That is the current setup.

adamnovak:
It looks like https://github.com/vgteam/sequenceTubeMap/commit/4cbfd1e488736051548d1affd05d022317eaac02 which adds the catch isn't actually in this branch.

adamnovak:
I merged everything in here so we actually have the catch and it seems to work. I think @rmccrear is indeed right about the end to end tests *only* bringing up the API server and not the static server.

@shreyasun I was also getting end-to-end test failures due to an outdated Node at one point; I was on Node 14 which was previously sufficient, but we now need Node v18.7.0 as specified in the `.nvmrc` file. So maybe make sure your Node is new enough also? If you have `nvm` installed you can `nvm install` and `nvm use` from the root of the repo to get the right Node in your current shell.

adamnovak:
How easy is Playwright to run? Right now `npm install` can get all our dependencies; would we need to require people to install a browser through another method or enable some kind of accessibility hooks on their machine to let it be controlled? And would it work well under Github Actions?

rmccrear:
It installs an isolated chromium browser with npm into the `node_modules` folder which then it controls with its own driver. (And can install other browsers, too.) So it doesn't use the user's browser or any special OS tools. It would usually run headless, but can be configured to run non-headless so you can see it and debug. And it does works under GitHub Actions, as well, since Actions are just run in Ubuntu instances and it would install a (headless) browser in the `node_module` on the instance, too. 

If you are considering it, I could put it into a separate branch or send you a link to it running on my fork if you wanted to see what it looks like first, and decide later.

PR #261, opened 2023-04-26
Install a compact test reporter for jest

rmccrear:
This PR just adds `jest-compact-reporter` as a reporter in addition to the default reporter.

It will cause `npm run test` to output Jest's default report followed by a summary of which tests passed and failed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "websocket": "^1.0.34"
       },
       "devDependencies": {
+        "jest-compact-reporter": "^1.2.9",
         "jest-fetch-mock": "^3.0.3",
         "prettier": "^2.7.1",
         "qs": "^6.11.0"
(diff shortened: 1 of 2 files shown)


adamnovak:
This seems to have broken saying `npm run test -- src/whatever-file.test.js` to run a single file of tests. It looks like the `--reporters` option grabs the test file name as one of its arguments.

@rmccrear Do you have any idea what to do about that?

rmccrear:
We could have two test commands: `ci:test` and `test`. Then `ci:test` could be run in the `.github/workflows/test.yml` file. And `test` could be reverted back.

Something like:

    // package.json
    "test": "react-scripts test",
    "ci:test": "react-scripts test --reporters default --reporters jest-compact-reporter",

    // .github/workflows/test.yml
    - name: Run tests
      run: PATH=$PATH:$(pwd)/bin CI=true npm run ci:test

PR #263, opened 2023-05-02
fix for test reporters

rmccrear:
The last merge broke the cli command `npm run test -- ./src/whatever.test.js`.

This fixes the problem by reverting `test` and creating a new command `ci:test` to be run in the CI environment on GitHub.

`ci:test` will run with the custom test reporters, while `test` will run as it did before.

This fixes the issue mentioned in https://github.com/vgteam/sequenceTubeMap/pull/261#issuecomment-1532042284

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,13 +36,6 @@ jobs:
           chmod +x bin/vg
       - name: Run tests
         run: PATH=$PATH:$(pwd)/bin CI=true npm run ci:test
-      - name: Produce Test Report
-        uses: dorny/test-reporter@v1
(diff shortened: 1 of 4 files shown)


rmccrear:
I also tried adding some nice html reporting with junit reporter (It looks like: [this](https://github.com/rmccrear/sequenceTubeMap/actions/runs/4908811716/jobs/8764832914]).)

But I found that it requires write permissions when running the action in order to save the report for display. So, I reverted that commit.

adamnovak:
I feel like there's got to be an option we can add to the end of that command line that will make it stop parsing --reporters and accept more options afterward.

PR #265, opened 2023-05-05
Track list

ducku:
- Add keys to radio row component
- Attach trackID to end of the testing tag to select specific items in the tracklist
- Rework TrackListItem to no longer rely on an event handler reference
    - Now keep a local version of uncommitted differences 
- Implement TrackList

### DIFF
--- a/src/components/TrackList.js
+++ b/src/components/TrackList.js
@@ -4,83 +4,66 @@ import {
     Row
   } from "reactstrap";
 import {TrackListItem} from './TrackListItem';
-import React, { useState, useRef, useEffect} from 'react';
 
 
 export const TrackList = ({
-    // tracks expects an object mapping an trackID to trackProps, which includes
-        // * trackType: string
-        // * trackFile: file object / undefined
-        // * trackColorSettings: object(aka. colorScheme)
-    tracks,
-    // availableTracks: array of tracks(see types.ts)
-    availableTracks,
-    // availableColors: array of ColorPalletes
-    availableColors,
-    onChange, // expects a new tracks object
-    onDelete,
-  }) => {
-    const [myTracks, setTracks] = useState(tracks);
-    const _onChange = useRef(onChange);
+  // tracks expects an object mapping an trackID to trackProps, which includes
+      // * trackType: string
+      // * trackFile: file object / undefined
+      // * trackColorSettings: object(aka. colorScheme)
+  tracks,
+  // availableTracks: array of tracks(see types.ts)
+  availableTracks,
+  // availableColors: array of ColorPalletes
+  availableColors,
+  onChange, // expects a new tracks object
+  onDelete,
+}) => {
 
-    function trackItemOnChange(trackID, trackProps) {
-        let newTracks = {...myTracks};
+  function trackItemOnChange(trackID, trackProps) {
+    (function(callback){ console.log(callback.toString()); })(onChange);
+      let newTracks = {...tracks};
+      newTracks[trackID] = trackProps;
+      if (JSON.stringify(newTracks) !== JSON.stringify(tracks)) {
+          onChange(newTracks);
+      }
 
-        newTracks[trackID] = trackProps;
-        if (JSON.stringify(newTracks) !== JSON.stringify(myTracks)) {
-            setTracks(newTracks);
-        }
-
-    }
-
-    useEffect(() => {
-        if (JSON.stringify(myTracks) !== JSON.stringify(tracks)) {
-          _onChange.current(myTracks);
-        }
-        
-      }, [myTracks, _onChange, tracks]);
-  
-
-
-    function renderTracks() {
-        let trackMarkdown = [];
-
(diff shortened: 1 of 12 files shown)


rmccrear:
In the test in `TrackSettingsButton.test.js`, it looks like it's not finding the button with `getByTestId('button')`

The aria-label for the button is "Settings", so you might try `getByRole('button', {name: 'Settings'})` instead.

src/components/TrackList.js
```diff
@@ -0,0 +1,70 @@
...
+      newTracks[trackID] = trackProps;
+      if (JSON.stringify(newTracks) !== JSON.stringify(tracks)) {
+          onChange(newTracks);
+      }
+
+  }
+
+
+  function renderTracks() {
+      let trackMarkdown = [];
```
adamnovak, line 35: There isn't any [Markdown](https://en.wikipedia.org/wiki/Markdown) here, right?

src/components/TrackList.test.js
```diff
@@ -0,0 +1,168 @@
...
+        fireEvent.click(getByText("haplotype"));
+
+        // should wait for file to be selected
+        expect(fakeOnChange).toHaveBeenCalledTimes(0);
+
+        // simulate dropdown effect
+        let newTracks = {...tracks};
+        newTracks[1].trackType = "haplotype";
+
+        rerenderTrackList(rerender, newTracks, fakeOnChange, fakeOnDelete);
```
adamnovak, line 108: Why is this step needed? The `TrackListItem` component is supposed to take care of holding its own state for how haplotypes has been selected but a file hasn't yet been picked, right? In the real page the track type in the list won't have changes yet, so why change it in the test?

src/components/TrackList.test.js
```diff
@@ -0,0 +1,168 @@
...
+                availableColors={availableColors}
+                onChange={fakeOnChange1}
+                onDelete={fakeOnDelete}
+            />
+        );
+
+        expect(fakeOnChange1).toHaveBeenCalledTimes(0); 
+
+        let newTracks = {...tracks};
+        newTracks[1].trackType = "graph"
```
adamnovak, line 154: This doesn't do anything, right? The track type for that entry is already `"graph"`.

src/components/TrackListItem.js
```diff
@@ -23,77 +23,81 @@ export const TrackListItem = ({
...
-    }, [myTrackProps.trackFile, myTrackProps.trackType, myTrackProps.trackColorSettings, myTrackProps, _onChange, trackProps]);
+      // push it if trackFile has been selected and changes are made
+      if (newTrackProps.trackFile !== undefined && JSON.stringify(trackProps) !== JSON.stringify(newTrackProps)) {
+        onChange(trackID, newTrackProps);
+        setPropChanges({});
+      }
+    }, [propChanges, onChange, trackProps, trackID]);
 
+    // displayed elements uses propChanges(local state) first, then uses trackProps
+    const displayedFile = "trackFile" in propChanges ? propChanges["trackFile"] : trackProps["trackFile"];
```
adamnovak, line 73: This isn't just an `||` because `propChanges` is supposed to be able to make the file undefined, right?

src/components/TrackListItem.js
```diff
@@ -23,77 +23,81 @@ export const TrackListItem = ({
...
+      setPropChanges(newPropChanges);
     }
 
     const trackSettingsOnChange = async(key, value) => {
       let newTrackColorSettings = {...trackProps["trackColorSettings"]};
       newTrackColorSettings[key] = value;
-      dispatch({type: "setColorSettings", payload: newTrackColorSettings});
+      let newPropChanges = {...propChanges};
+      newPropChanges["trackColorSettings"] = newTrackColorSettings;
+      setPropChanges(newPropChanges);
```
adamnovak, line 52: Is it really better to set component state here and then have it re-render with the new state and then have useEffect see the state is safe to commit upwards and then have it call `onChange` and change the props and state again and then have it render again?

I feel like each event handler could call a `commitOrBufferPropChanges()` function with `newPropChanges`, and right inside the event handler we could decide if we are ready to call up to `onChange` or if we need to buffer the changes in `propChanges`, and that might be simpler and could save some renders.

But this looks like it ought to work the way it is.

adamnovak (changes requested):
I noticed one variable name that I think is wrong, and I'm not sure that all the steps in the tests are justified. After those are fixed I think this is good to merge! Thanks!

src/components/TrackListItem.js
ducku: I was going down this path, but I ran into an issue where calling a setState hook and using its value afterward doesn't guarantee coherence. I'm not sure if I like the use of useEffect either, but it seemed like the consensus is that this works.

https://stackoverflow.com/questions/54069253/the-usestate-set-method-is-not-reflecting-a-change-immediately
https://stackoverflow.com/questions/54954091/how-to-use-callback-with-usestate-hook-in-react

adamnovak (approved):

PR #271, opened 2023-05-10
Add --seed to terminate --reporters parsing

adamnovak:
This fixes:
```
npm run test -- src/components/HelpButton.test.js
```

It adds a no-op option (setting `--watchAll` to its default) to stop `--reporters` eating the positional arguments.

You can still override it the normal way, because the last option wins:
```
npm run test -- --watchAll=false src/components/HelpButton.test.js
```

I think this will close https://github.com/vgteam/sequenceTubeMap/pull/263 because now we can just use `npm run test` all the time and always have the nice summary reporter, and we won't need a separate `npm run ci:test`.


rmccrear:
Yes, this works well, and it can replace #263.

PR #275, opened 2023-05-15
Single color gams/haplotype

ducku:
- Implement ColorPicker component that uses React-Color to pick a single color
- Integrate ColorPicker into TrackSettings
- Allow tubemap to take a React-Color object in addition to a set of colors

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -214,18 +214,8 @@ input[type="file"] {
   right: -20px;
 }
 
-/*
-.col-md-auto{
-  width: 70%;
-}
-
-
-.form {
-  max-width: 100%
-}
-*/
-
 .radio-row {
-  max-width: 52%;
+  /*max-width: max-content;  for some reason doesn't work for haplotype settings row */
+  max-width: 680px; /* TODO: would like to make this flexible? */
   margin-top: 10px;
 }
--- a/src/App.css
+++ b/src/App.css
@@ -213,3 +213,19 @@ input[type="file"] {
   top: -20px;
   right: -20px;
 }
+
+/*
+.col-md-auto{
+  width: 70%;
+}
+
+
+.form {
+  max-width: 100%
+}
+*/
+
+.radio-row {
+  max-width: 52%;
+  margin-top: 10px;
+}
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -64,19 +64,19 @@ enum ColorPalette{
 }
 
 // Hex describing a color
-type colorHEX = `#${string}`;
+type ColorHex  = `#${string}`;
 
 
 // An object created by the React-Color library, contains infomation on the color selected
-type React_Color = {
-  hex: colorHEX,
+type ReactColor = {
+  hex: ColorHex ,
   //... has additional properties
 }
 
 // Describes the coloring information for a track
 type ColorScheme = {
-    mainPalette: ColorPalette | React_Color,
-    auxPalette: ColorPalette | React_Color,
+    mainPalette: ColorPalette | ColorHex ,
+    auxPalette: ColorPalette | ColorHex ,
     colorReadsByMappingQuality: boolean
 }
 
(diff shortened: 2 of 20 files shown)


src/App.css
```diff
@@ -213,3 +213,19 @@ input[type="file"] {
...
+/*
+.col-md-auto{
+  width: 70%;
+}
+
+
+.form {
+  max-width: 100%
+}
+*/
```
adamnovak, line 226:
```suggestion
```

This looks unused.

src/Types.ts
```diff
@@ -54,19 +54,29 @@ type RegionInfo = {
...
+type React_Color = {
+  hex: colorHEX,
+  //... has additional properties
 }
 
 // Describes the coloring information for a track
 type ColorScheme = {
-    mainPallete: ColorPallete,
-    auxPallete: ColorPallete,
+    mainPalette: ColorPalette | React_Color,
```
adamnovak, line 78: Why not just use `colorHEX` instead of a whole `React_Color`?

src/Types.ts
```diff
@@ -54,19 +54,29 @@ type RegionInfo = {
...
   ygreys,
   blues,
   reds,
   plainColors,
-  lightColors
+  lightColors,
+}
+
+// Hex describing a color
+type colorHEX = `#${string}`;
```
adamnovak, line 67: Maybe this should be called something like `ColorHexCode` or `ColorHex` to match the formatting of the other types?

src/util/tubemap.js
```diff
@@ -2315,6 +2315,13 @@ function calculateTrackWidth() {
 
 
 function getColorSet(colorSetName) {
+  // could be a color object from react-color
+  if(typeof colorSetName == "object") {
```
adamnovak, line 2319: Oh I guess you are using an object or a string as the way you determine whether you want to use a hex code or a defined palette?

I suppose you could use whether the string starts with "#" instead, though, right?

adamnovak (changes requested):
This looks like it ought to work, but I found some things that I think should be renamed, and I'm not sure using a custom dialog and a whole color picker package is giving us any advantage over `<input type="color">`, which should work fine in React like all the other browser builtins do.

Also, if you pick a palette and then want to go back to a custom color, how do you do that? Do you have to re-pick it? Does it get lost from the UI?

ducku:
Here's a screenshot of the sketch color picker. 
Currently, it updates the tubemap visuals upon exiting the popup. I'm not sure if it's the most intuitive.

image

adamnovak:
> Currently, it updates the tubemap visuals upon exiting the popup. I'm not sure if it's the most intuitive.

I think that makes sense, as long as the user actually selected something in the popup.

It would be nice to get the tube map itself to re-render more live, but I think changing the colors also like resets the zoom right now, which we might want to fix first.

src/Types.ts
```diff
@@ -54,19 +54,29 @@ type RegionInfo = {
...
+
+// Hex describing a color
+type ColorHex  = `#${string}`;
+
+
+// An object created by the React-Color library, contains infomation on the color selected
+type ReactColor = {
+  hex: ColorHex ,
+  //... has additional properties
 }
```
adamnovak, line 74: This might not be used anymore?

src/components/ColorPicker.js
```diff
@@ -0,0 +1,62 @@
...
+    position: 'fixed',
+    top: '0px',
+    right: '0px',
+    bottom: '0px',
+    left: '0px',
+    }
+
+    return (
+        <div>
+            <Button aria-label="ColorPicker" onClick={() => setOpen(!open)} data-testid={testID}><FontAwesomeIcon icon={faPalette} /></Button>
```
adamnovak, line 33: The `aria-label` could be better here. I don't know if the non-dictionary-word "ColorPicker" is a good choice, because being a non-word it might confuse not-very-clever speech-to-text/text-to-speech setups?

adamnovak (approved):
This looks like it should work, and it has a screenshot that looks good.

It could be improved with unit tests.

Issue #259, opened 2023-04-26
Handle commas in coordinates


adamnovak:
Right now, if you have thousands-separator commans in your region specification, they are passed directly to vg. vg then parses the number with a parser that stops at the first non-digit, so each coordinate parses as just the part up to the first thousands separator. So like in this screenshot I can ask for a few megabases and get a few bases:

![Image](https://user-images.githubusercontent.com/5062495/234686756-39e318e3-6e8d-43f2-86c4-1132736f5db4.png)

We should drop the commas from the numbers before sending the range to vg.

PR #278, opened 2023-05-18
Handle commas in coordinates

shreyasun:
Handle commas in coordinates #259


adamnovak (approved):

Issue #228, opened 2023-01-31
Implement `TrackPickerDisplay`


adamnovak:
Create a `TrackPickerDisplay` component, as defined in #205. It should consist of a `TrackList` and a `TrackAddButton`. It should take as props `tracks`, `availableTracks`, and `onChange`, as defined for `TrackList` in #226. It should call `onChange` when a track is modified in the `TrackList` or when a new track is added via the `TrackAddButton`.

Similar to `TrackListItem`, this component might need to keep some of its own internal state to e.g. let you click the add button and get a `TrackListItem` in an incompletely-filled state, with the call to `onChange` being delayed until the user has finished selecting what they want the new track to be. Alternatively, the new track could be added in some arbitrary state, as long as the type is appropriate to the filename, and it is compatible with the existing tracks. (For example, there cannot be two "graph" tracks at the same time.)

Issue #227, opened 2023-01-31
Implement `TrackAddButton`


adamnovak:
Create a `TrackAddButton` control as defined in #205. This should be a wide button with a "+" icon and appropriate tooltips and accessibility properties. When clicked, it should call the function in its `onClick` prop.

PR #279, opened 2023-05-21
Implement TrackListDisplay

ducku:
Issues #227 and #228

Implement TrackListDisplay, with functionality for add and delete buttons. 

- Local state of Track Items is kept
       - Changes are pushed when all files have been selected
- Add and delete buttons modify the track list respectively

### DIFF
--- a/src/components/TrackList.test.js
+++ b/src/components/TrackList.test.js
@@ -98,8 +98,10 @@ describe('TrackList', () => {
         await waitFor(() => getByText("haplotype"));
         fireEvent.click(getByText("haplotype"));
 
-        // should wait for file to be selected
-        expect(fakeOnChange).toHaveBeenCalledTimes(0);
+        expect(fakeOnChange).toHaveBeenCalledTimes(1);
+        let newTracks = JSON.parse(JSON.stringify(tracks));
+        newTracks[1].trackType = "haplotype";
+        rerenderTrackList(rerender, newTracks, fakeOnChange, fakeOnDelete);
 
         // change file name
         const fileSelectComponent = queryByTestId('file-select-component1');
@@ -107,9 +109,8 @@ describe('TrackList', () => {
         await waitFor(() => getByText("fileB1.gbwt"));
         fireEvent.click(getByText("fileB1.gbwt"));
 
-        expect(fakeOnChange).toHaveBeenCalledTimes(1);
-        let newTracks = JSON.parse(JSON.stringify(tracks));
-        newTracks[1].trackType = "haplotype";
+        expect(fakeOnChange).toHaveBeenCalledTimes(2);
+        //newTracks[1].trackType = "haplotype";
         newTracks[1].trackFile = {"name": "fileB1.gbwt", "type": "haplotype"};
         expect(fakeOnChange).toHaveBeenCalledWith(newTracks);
 
@@ -123,7 +124,7 @@ describe('TrackList', () => {
         fireEvent.click(getByText("reds"));
         fireEvent.click(document);
 
-        expect(fakeOnChange).toHaveBeenCalledTimes(2); 
+        expect(fakeOnChange).toHaveBeenCalledTimes(3); 
 
         newTracks[1].trackColorSettings.mainPalette = "reds";
         expect(fakeOnChange).toHaveBeenCalledWith(newTracks);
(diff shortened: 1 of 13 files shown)


src/components/TrackPickerDisplay.js
```diff
@@ -0,0 +1,104 @@
...
+            const trackProps = myTrackList[trackID];
+            if (trackProps.trackFile === undefined) {
+                validTrackList = false;
+            }
+        });
+
+        // commit changes when files are valid and changes are made
+        if (validTrackList && JSON.stringify(tracks) !== JSON.stringify(myTrackList)) {
+            onChange(myTrackList);
+        }
```
adamnovak, line 72: I think you need to clear out `myTrackList` here so we don't constantly try and clobber whatever is being sent down. We only want to try to apply any given user edits once.

adamnovak (changes requested):
This looks pretty good!

I'm worried it has the same constantly-reapply-local-state problem that the TrackListItem had, since I don't see where `newTrackList` gets cleared.

I'm also a little worried that we're starting to properly deviate from my notion of the track list as a real *list* (i.e. an array numbered 0, 1, 2, ... with no gaps). We can do it this way, but we might want to rename some stuff (like the track list) if it's really a non-list-shaped collection. Or we can make it a real denser array list and use `slice` to cut things out of it, and then deal with assigning the items unique keys for the React key system.

src/components/TrackPickerDisplay.js
```diff
@@ -0,0 +1,104 @@
...
+            newTrackList[trackID] = newTracks[trackID];
+        })
+        setMyTrackList(newTrackList);
+    }
+
+    const onDelete = (trackID) => {
+        let newTrackList = {...myTrackList};
+        delete newTrackList[trackID];
+        setMyTrackList(newTrackList);
+    }
```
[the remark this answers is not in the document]
adamnovak: Also, is the iteration order to render the track list guaranteed to be in order of increasing numerical key value? Or if you add 10 tracks will track `"10"` sort after track `"1"` and before track `"2"`?

src/components/TrackList.test.js
```diff
@@ -98,18 +98,19 @@ describe('TrackList', () => {
...
         const fileSelectComponent = queryByTestId('file-select-component1');
         fireEvent.keyDown(fileSelectComponent.firstChild, {key: "ArrowDown"});
         await waitFor(() => getByText("fileB1.gbwt"));
         fireEvent.click(getByText("fileB1.gbwt"));
 
-        expect(fakeOnChange).toHaveBeenCalledTimes(1);
-        let newTracks = JSON.parse(JSON.stringify(tracks));
-        newTracks[1].trackType = "haplotype";
+        expect(fakeOnChange).toHaveBeenCalledTimes(2);
+        //newTracks[1].trackType = "haplotype";
```
adamnovak, line 113:
```suggestion
```

src/components/TrackPickerDisplay.js
```diff
@@ -0,0 +1,120 @@
...
+import {
+    Row,
+    Col
+  } from "reactstrap";
+import {TrackList} from "./TrackList";
+import {TrackAddButton} from "./TrackAddButton";
+import config from "./../config.json";
+
+export const TrackPickerDisplay = ({
+    // tracks expects an object mapping an trackID to trackProps, which includes
```
adamnovak, line 12:
```suggestion
    // tracks expects an object mapping a trackID to trackProps, which includes
```

src/components/TrackPickerDisplay.js
```diff
@@ -0,0 +1,120 @@
...
+        setTrackListChanges(newTrackList);
+    }
+
+
+    useEffect(() => {
+        // construct the new track list
+        let newTrackList = applyTrackListChanges(tracks, trackListChanges)
+
+        let validTrackList = true;
+        // track list is valid to commit if all fileNames have been selected
```
adamnovak, line 71: This is going to mean that changes to one track won't commit as long as another track exists that isn't fully filled in, right?

We might need some way to guide the user towards that offending invalid track, or they might be confused if the change something in another track, hit go, and don't see it take effect.

adamnovak (approved):
We might want to do something else to make it clear to the user when their changes are and are not going through, if we can only propagate changes up when all items are ready. But maybe we can lump that in with other validation for stuff like making sure a graph track exists.

Issue #242, opened 2023-03-15
Link to view URLs do not populate dropdowns with correct values when followed


adamnovak:
I can:

1. Start up the tube map
2. Set it to custom (mounted files) mode
3. Set graph to `cactus.vg`
4. Set region to `ref:1-10`
5. Click `Go`
6. Click `Copy link to data`

Then I get a link like `http://127.0.0.1:3001?tracks[0][files][0][name]=cactus.vg&tracks[0][files][0][type]=graph&tracks[1][files][0][type]=haplotype&tracks[2][files][0][type]=read&tracks[3][files][0][type]=read&bedFile=none&region=ref:1-10&dataPath=mounted&dataType=mounted files`.

If I paste that into my address bar in a new tab, I get the right view displayed, but the graph file dropdown is set to `none`, and not `cactus.vg`.

I think this has something to do with the track system refactor in and the duplication of state between the new system and the old one, which has not yet been fully resolved.

Going to the link should set the view *and* set the dropdowns.

We should add a new test to enforce this, and fix it.

PR #280, opened 2023-05-25
Use track list as the source of truth about what files are selected, and avoid empty tracks

adamnovak:
This removes the dedicated variables for what files are selected in each dropdown, and instead uses an array of tracks.

It also makes graph/haplotype paths use a `sourceTrackID` like reads do.

It also removes the assumptions that the graph is track 0, the haplotypes if any are track 1, and the reads are tracks 2 and 3. Any file type should be able to be any track number.

This is all in service of making following a copied link actually populate the file dropdowns. When you visit a copied link, it fills in the view target's tracks. Now we use that information to render the dropdowns, so #242 should be fixed.


src/App.js
```diff
@@ -14,6 +14,33 @@ import Footer from "./components/Footer";
...
+      if (tracks[key].files[0].type === "read") {
+        schemes[key] = {...config.defaultReadColorPalette};
+      } else {
+        schemes[key] = {...config.defaultHaplotypeColorPalette};
+      }
+    }
+  }
+  
+  return schemes;
+}
```
ducku, line 42:
```suggestion
function addTracksToColorSchemes(tracks, schemes) {

  console.log("Initializing coloring for tracks: ", tracks)

  for (const key in tracks) {
    if (schemes[key] === undefined) {
      // We need to adopt a color scheme
      if (tracks[key].files[0].type === "read") {
        schemes[key] = {...config.defaultReadColorPalette};
      } else {
        schemes[key] = {...config.defaultHaplotypeColorPalette};
      }
    }
  }

  return schemes;
}
```

src/App.js
```diff
@@ -71,9 +94,34 @@ class App extends Component {
...
+        for (const key in newViewTarget.tracks) {
+          if (newColorSchemes[key] === undefined) {
+            // We need to adopt a color scheme
+            if (newViewTarget.tracks[key].files[0].type === "read") {
+              newColorSchemes[key] = {...config.defaultReadColorPalette};
+            } else {
+              newColorSchemes[key] = {...config.defaultHaplotypeColorPalette};
+            }
+          }
+        }
```
ducku, line 113:
```suggestion
        let newColorSchemes = state.visOptions.colorSchemes.slice();
        addTracksToColorSchemes(newViewTarget.tracks, newColorSchemes);

```

src/components/TubeMapContainer.js
```diff
@@ -141,36 +141,52 @@ class TubeMapContainer extends Component {
...
-        // Go through every read and assign it a source file number
-        for (let i = 0; i < readsArr.length; i++) {
-          for (let j = 0; j < readsArr[i].length; j++) {
-            readsArr[i][j].sourceTrackID = sourceTrackIDs[i];
-          }
-        }
+        console.log("Graph track: " + graphTrackID + " Haplotype track: " + haplotypeTrackID);
+
+        const nodes = tubeMap.vgExtractNodes(json.graph);
+        // TODO: For now claim haplotypes came from the graph
```
ducku, line 174: I'm not sure if I understand what needs to be done here. Do haplotypes need to be extracted separately?

src/components/HeaderForm.js
```diff
@@ -169,6 +194,79 @@ class HeaderForm extends Component {
...
+        if (track.files[0].type === type) {
+          console.log("See file " + seenTracksOfType + " of right type");
+          if (seenTracksOfType === index) {
+            if (file !== "none") {
+              // We want to adjust it, so keep a modified copy of it
+              let newTrack = JSON.parse(JSON.stringify(track));
+              newTrack.files[0].name = file;
+              newTracks[key] = newTrack;
+            }
+            // If the file is "none" we drop the track.
```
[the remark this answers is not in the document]
adamnovak: Yeah, I think we want to compact them eventually. But I figured starting on that here wouldn't work well with what you are doing for the track picker list, which wants thinks keyed by stable IDs. I think we need to make that change after integrating.

PR #283, opened 2023-06-01
Read info click dialog

### LINKED ISSUE #273: Read info dialog that supports copy-paste of current tooltip info
We want to be able to click on reads and open dialogs about them that you can copy-paste from.

shreyasun:
Fixes #273


src/util/tubemap.js
```diff
@@ -3240,6 +3248,8 @@ function filterObjectByAttribute(attribute, value) {
   return (item) => item[attribute] === value;
 }
 
+// draw clicks
+
```
adamnovak, line 3252: I don't think that's right?
```suggestion
```

Or you could explain how this draws "track rectangles", which I think are the straight non-bendy parts of the reads and paths.

src/util/tubemap.js
```diff
@@ -3599,6 +3615,27 @@ function trackDoubleClick() {
...
+    index < inputTracks.length &&
+    inputTracks[index].id !== Number(trackID)
+  ) {
+    index += 1;
+  }
+  if (index >= inputTracks.length) return;
+  if (DEBUG) console.log(`moving index: ${index}`);
+  moveTrackToFirstPosition(index);
+  createTubeMap();
+  */
```
adamnovak, line 3636: We don't need this commented-out code probably.
```suggestion
```

src/components/PopUpTrackText.js
```diff
@@ -0,0 +1,51 @@
...
+PopUpTrackText.propTypes = {
+  open: PropTypes.bool.isRequired,
+  text: PropTypes.string.isRequired,
+  close: PropTypes.func.isRequired,
+}
+
+PopUpTrackText.defaultProps = {
+  open: PropTypes.string.isRequired,
+  close: PropTypes.func.isRequired,
+}
```
adamnovak, line 51: These are meant to be the default values to use. If a prop's type in `propTypes` has `.isRequired` on it, there's no sense in giving it a default value, since the default can never be used, because the prop will always have a value provided.

And if you do want to set default values, they have to be actual values (so like a boolean of `false` or `true` for `open`, and some function for `close`). They shouldn't themselves be the things in `PropTypes` that we use to name the types of things.

I think we could probably just cut `PopUpTrackText.defaultProps`.
```suggestion
```

adamnovak (changes requested):
I think this looks pretty good, but I made a suggestion that I think will fix CI and some others to clean up commented-out code and stuff that didn't make sense to me.

I think you can go in the "Files changed" view and add the suggestions to a batch and commit the batch, and Github will make a commit that takes them.

PR #285, opened 2023-06-02
Merge TrackPicker

ducku:
I decided to put the button component directly into TrackPicker, it made sense to me to skip a generic button component.

- TrackPicker calls onChange when TrackPickerDisplay does, perhaps it might be better to do this when the user exits the popup?
- Changed TrackSettings to exit upon clicking outside the popup, while the TrackPicker popup only exits upon clicking the X button
- TrackListItem now only takes up a fixed # of pixels in width, removing white space in TrackPicker

### DIFF
--- a/src/components/TrackPicker.test.js
+++ b/src/components/TrackPicker.test.js
@@ -169,10 +169,9 @@ describe('TrackPicker', () => {
 
     });    
 
-
-    it('should delete a trackitem when the delete button is pressed', async () => {
+    it('should close when the exit button is pressed', async () => {
         const fakeOnChange = jest.fn();
-        const { queryByTestId } = render(
+        const { queryByTestId, getByText, rerender } = render(
             <TrackPicker
                 tracks={tracks}
                 availableTracks={availableTracks}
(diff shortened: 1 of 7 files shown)


src/components/TrackPicker.test.js
```diff
@@ -0,0 +1,196 @@
...
+
+        expect(queryByTestId("settings-button-component1")).toBeTruthy();
+
+        expect(queryByTestId("delete-button-component1")).toBeTruthy();
+
+        expect(queryByTestId("track-add-button-component")).toBeTruthy();
+    });
+
+
+    it('should add track items when the add button is pressed', async () => {
```
adamnovak, line 52: It looks like this test and some of the others might be more duplicative than they need to be with the TrackPickerDisplay tests. We should be able to assume that the TrackPickerDisplay works already, and so e.g. clicking its add button will add a track. What we really care abut here is making sure the popup opens and closes, and making sure changed actually propagate through when made in the TrackPickerDisplay, and that they do so at the right time relative to opening/closing the dialog.

adamnovak (changes requested):
The code here looks pretty good; maybe some of the tests could be trimmed to avoid testing the TrackPickerDisplay as thoroughly here for a different component.

But I noticed there wasn't a screenshot, and that `1000px` is a pretty wide width if you might only get e.g. half a 1920px wide monitor for a browser window. So I checked it out to look at, and I'm not sure it is rendering right. It looks like this for me in Firefox:
TrackPicker rendering weirdly narrow

I feel like the width stuff is not working right.

In general we should prefer being responsive over trying to force a particular width. While people probably have a screen as wide as or wider than 1000 pixels, they might not want to dedicate all that width to the tube map; maybe they want to look at two tube maps side by side, or have the tube map open on one side of the screen while looking at the code responsible for a mismapped read on the other side. So we should try and let things get wider or narrower as space allows, with some maximum visually-appropriate width limit.

adamnovak:
I think part of what is going on with the width is that TrackListItem only uses 6 of Bootstrap's 12 columns in total, among its `Column` elements.

It might also be that `width: fit-content` on the container wants to make the container only wide enough to fit the content, while the Bootstrap column-based system wants to make each column some fraction of the width of the container. So without anything asking it to be some ideal width or some fraction of the browser viewport width everything becomes as small as it can be before the minimum widths of the controls and the space between them start to kick in.

adamnovak (approved):

PR #287, opened 2023-06-06
new popup dialog component

### LINKED ISSUE #267: Make the help dialog look like the track settings dialog
Now the track settings dialog has a nice shadow and close button, but the help dialog doesn't. They should be consistent and maybe use a common slightly-customized-dialog component.

shreyasun:
Fixes #267


adamnovak (approved):
This looks good to me. We presumably could use a demo or tests for this component separately, but I think it is pretty simple to inspect and it is exercised by other tests.

PR #292, opened 2023-06-12
Documentation on how to make and use docker container

jmonlong:
This is how I usually use the tubemap, on servers or locally. 

It works with the latest commit, but we might want to have a system of tag/version to tag the docker container too.

I've pushed an image on https://quay.io/repository/jmonlong/sequencetubemap for now.


adamnovak:
We probably don't want to use `--no-check-certificate` for the vg download if we can avoid it, but I guess we can fix that later. Maybe by basing off a vg release image or something.

Thanks for setting this all up!

PR #293, opened 2023-06-14
Integrate track picker

ducku:
Integrate track picker into the mounted files section

Notable Changes:
- The track data structure has been updated to support a single file rather than an array of files
- Tracks are stored in an object rather than an array, tracks are identified by an unique trackID

Bugs:
- Read coloring only partially works. Certain strands are not updated when the color changes.
- Region area does not display autocomplete until something has been typed


adamnovak:
> Region area does not display autocomplete until something has been typed

I don't think that's a new bug introduced here; I was seeing that as well without this PR; I've been having to hit the down arrow to pop it open.

src/App.js
```diff
@@ -21,14 +21,16 @@ const EXAMPLE_TRACKS = [
   {"files": [{"type": "read", "name": "fakeReads"}]}
 ];
 
-function addTracksToColorSchemes(tracks, schemes) {
+function getColorSchemesFromTrack(tracks) {
```
adamnovak, line 24: This gets all the schemes form all the tracks, so it should probably be `getColorSchemesFromTracks` plural.

src/Types.ts
```diff
@@ -22,7 +22,13 @@ type file = {
...
 type track = {
-  files: Array<file>
+  trackFile: file;
+  trackType: filetype;
+  trackColorSettings: ColorScheme;
+}
+
+type tracks = {
+  trackID: track;
 }
```
adamnovak, line 32: Is this how you specify a dict from string to `track` in TypeScript syntax?

src/Types.ts
```diff
@@ -22,7 +22,13 @@ type file = {
 
 // Contains information necessary to make a track
 type track = {
-  files: Array<file>
+  trackFile: file;
+  trackType: filetype;
+  trackColorSettings: ColorScheme;
+}
```
adamnovak, line 28: Now we have duplicate copies of the type for the file.

We should probably get rid of `file`, and just have `track` have a filename string and a track type.

Or, we could store the type only in the file and not in the track.

But we should not store it twice.

src/components/TubeMapContainer.js
```diff
@@ -65,6 +65,28 @@ class TubeMapContainer extends Component {
...
+    tubeMap.setColorReadsByMappingQualityFlag(visOptions.colorReadsByMappingQuality);
+
+    for (const trackID in visOptions.colorSchemes) {
+      console.log("setting color: ", trackID, visOptions.colorSchemes[trackID]);
+      tubeMap.setColorSet(trackID, visOptions.colorSchemes[trackID]);
+    }
+
+    tubeMap.setMappingQualityCutoff(visOptions.mappingQualityCutoff);
+  }
+
```
adamnovak, line 89: I think I moved this over to the `TubeMap` component; it looks like you are adding a new copy and not adding any calls to it.

Maybe something here is why the colors aren't applying properly?

src/util/tubemap.js
```diff
@@ -2350,6 +2349,10 @@ function getColorSet(colorSetName) {
 
 function generateTrackColor(track, highlight) {
 
+    
+  //console.log("color schemes", config.colorSchemes);
+  //console.log("source ID", track.sourceTrackID);
```
adamnovak, line 2354:
```suggestion
```

We should cut out this commented-out code.

adamnovak (changes requested):
This looks pretty good!

I found some stuff that could be renamed/refactored, and I think I might have a vague lead on why the colors aren't applying right.

adamnovak:
To put the log in a file you can do:
```
npm run test -- --watchAll=false test-names-if-you-want-any >log.txt 2>&1
```

ducku:
Bug:

When selecting a singular color from the color picker, not all reads are swapped to that color.

I've selected green as the read color on a singular read file, but some reads are still the default read color.
image

adamnovak:
It looks like the colors still don't update live, though the changes to the track design look good.

Do we want to change it to make people press the button to update the colors?

src/end-to-end.test.js
```diff
@@ -36,12 +36,14 @@ let root = undefined;
 // Mock clipboard (string)
 let fakeClipboard = undefined;
 
+let dom;
```
adamnovak, line 39: Don't we not read this ever?

adamnovak (changes requested):
This looks pretty good, although I think I found some dead code that ought to be cut. I will see if I can do that later today, or @ducku you can try to do it first.

adamnovak (approved):

PR #298, opened 2023-06-21
Fixes #281

### LINKED ISSUE #281: Work around Google Chrome not providing the clipboard API
As noted in #274, if the tube map is accessed over plain-text HTTP and is running somewhere other than `localhost`, Google Chrome doesn't implement the clipboard API. Since the tube map is meant to work in a lot of academic environments, we expect people to e.g. run it on one end of a VPN connection on a machine that can't really get an Internet-facing HTTPS certificate.

We should add some kind of workaround to still let people get ahold of the link to the current view, even if the clipboard API is not available.

We could also work out a good way to set up the tube map on HTTPS with a self-signed cert, and turn that on by default.

shreyasun:
Fixes #281

### DIFF
--- a/src/components/CopyLink.js
+++ b/src/components/CopyLink.js
@@ -25,8 +25,6 @@ export function CopyLink(props) {
   const [dialogLink, setDialogLink] = useState(undefined);
 
   const handleCopyLink = () => {
-    // open popup
-    setOpen(!open);
     // Turn viewTarget into a URL query string
     const viewTarget = props.getCurrentViewTarget();
     // Don't stringify objects for readability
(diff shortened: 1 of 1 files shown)


src/components/CopyLink.js
```diff
@@ -3,43 +3,68 @@ import { Button } from "reactstrap";
...
+
 export function CopyLink(props) {
   // Button to copy a link with viewTarget to the data selected
-
   const [text, setText] = useState(UNCLICKED_TEXT);
+  const [dialogLink, setDialogLink] = useState(undefined);
 
   const handleCopyLink = () => {
+    // open popup
+    setOpen(!open);
```
adamnovak, line 29: Isn't this going to make the dialog always open, even if `navigator.clipboard` works fine?

adamnovak (changes requested):
This looks pretty good, except I think the dialog is going to open even when it isn't needed.

adamnovak (approved):
This looks like it should work great!

PR #300, opened 2023-06-27
Modified encoding url specifications to encodeValuesOnly: true to ensure proper url encoding for link to work correctly. Also modified tests to correctly expect the modified url.

### LINKED ISSUE #241: Link to view URLs contain unescaped spaces
For a custom (mounted files) view, I can click the link to view button and get a URL like this in my clipboard:

```
http://127.0.0.1:3001?tracks[0][files][0][name]=part.vg&tracks[0][files][0][type]=graph&tracks[1][files][0][name]=none&tracks[1][files][0][type]=haplotype&tracks[2][files][0][name]=sim.sorted.gam&tracks[2][files][0][type]=read&tracks[3][files][0][name]=none&tracks[3][files][0][type]=read&bedFile=none&region=GRCh38.chr14:1-10&dataPath=mounted&dataType=mounted files
```

The space in `mounted files` is not escaped, so this URL can't be pasted into Github and auto-linked, or used as the destination of a Markdown link, unless the space is first replaced with `%20`. We should make sure to properly urlencode all values here.

shreyasun:
Fixes #241

### DIFF
--- a/src/components/CopyLink.js
+++ b/src/components/CopyLink.js
@@ -31,7 +31,7 @@ export function CopyLink(props) {
     const viewTarget = props.getCurrentViewTarget();
     // Don't stringify objects for readability
     // https://github.com/ljharb/qs#stringifying
-    const params = qs.stringify(viewTarget, { encode: false });
+    const params = qs.stringify(viewTarget, { encodeValuesOnly: true });
     // complete url
     const full = window.location.origin + "?" + params;
     console.log(full);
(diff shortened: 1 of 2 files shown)


adamnovak (approved):
This looks great!

I'm going to merge it after the track picker stuff though because I anticipate a merge conflict on the correct URLs in the tests and I think it will be simpler to resolve that way.

PR #307, opened 2023-07-06
More read info dialog

### LINKED ISSUE #294: Add sample name/read group and CIGAR string to read info dialog

shreyasun:
Fixes #294


src/components/PopUpTrackInfo.js
```diff
@@ -0,0 +1,41 @@
...
+}) => {
+    return(
+      <div>
+        <PopupDialog open={open} close={close}>
+            <h5>Object Information</h5>
+            <table>
+              {/* Node info here */}
+              {(attributes || []).map(function(attribute){
+                return <tr key={attribute[0]} style={{border: '1px solid black'}}>
+                  <td style={{border: '1px solid black', fontWeight: 'bold'}}>{attribute[0]}</td>
```
adamnovak, line 18: Having a border on the row and also on one cell is a little weird, but it seems to look right.

src/util/tubemap.js
```diff
@@ -3702,10 +3703,22 @@ function trackDoubleClick() {
 
 function trackSingleClick() {
   /* jshint validthis: true */
-  const trackName = d3.select(this).attr("trackName");
+  const trackID = d3.select(this).attr("trackID");
+  let current_track = tracks[trackID];
```
adamnovak, line 3707: Something is wrong here somehow; when I click on a read I get a popup for a *different* read, with the name and CIGAR and other info for some other read. I think maybe we are doing something like sorting the reads between when the "trackID" D3 attribute is set and when we read it out and get the track? Or maybe what we put in isn't the right thing?

src/util/tubemap.js
```diff
@@ -3919,6 +3932,73 @@ function compareReadsByLeftEnd2(a, b) {
   return 0;
 }
 
+// converts readPath to a CIGAR string
```
adamnovak, line 3935: What is `readPath` meant to be? I think it's a vg Path object expressed as a JS object, right?

src/util/tubemap.test.js
```diff
@@ -0,0 +1,214 @@
...
+              }
+            ]
+          },
+          {
+            edit: [
+              {
+                from_length: 0,
+                sequence: "C"
+              }
+            ]
```
adamnovak, line 208: This doesn't actually make any sense, since if sequence is set its length always has to be `to_length`. But I suppose it is a good test case to see if we can break it.

adamnovak (changes requested):
This looks good except I think it needs to display the right read and not the wrong read when you click a read.

src/components/PopUpTrackInfo.js
```diff
@@ -0,0 +1,41 @@
...
+  close,
+}) => {
+    return(
+      <div>
+        <PopupDialog open={open} close={close}>
+            <h5>Object Information</h5>
+            <table>
+              {/* Node info here */}
+              {(attributes || []).map(function(attribute){
+                return <tr key={attribute[0]} style={{border: '1px solid black'}}>
```
adamnovak, line 17: It looks like React doesn't like simple tables without a `<tbody>` tag. I'm getting it throwing up a warning in the console when the table renders:

```
Warning: validateDOMNesting(...): <tr> cannot appear as a child of <table>. Add a <tbody>, <thead> or <tfoot> to your code to match the DOM tree generated by the browser.
tr
table
div
div
CardBody@http://127.0.0.1:3001/static/js/bundle.js:148921:19
div
Card@http://127.0.0.1:3001/static/js/bundle.js:148822:19
div
Container@http://127.0.0.1:3001/static/js/bundle.js:151747:19
div
div
./node_modules/reactjs-popup/dist/reactjs-popup.esm.js/Popup<@http://127.0.0.1:3001/static/js/bundle.js:146711:22
div
PopupDialog@http://127.0.0.1:3001/static/js/bundle.js:3707:14
div
```

It looks like the browser automatically creates a `<tbody>` if the table doesn't have one, and React demands that you not let it do that and actually create the `<tbody>` manually instead, so that the browser isn't adding markup that your code doesn't know about.

We had better do what Facebook says I guess, I don't think there's a way to turn that warning off.

src/util/tubemap.js
adamnovak: @shreyasun I think I fixed this by just brute force scanning for the right track.

adamnovak (approved):
I had to fix the ID comparison since changing to `===` broke it, since it was actually matching numbers against strings.

I also fixed the `<tbody>` setup; you are supposed to have *one* `<tbody>` for the whole table and not one per row, and also having each row have a `<tbody>` but having the `key` attribute on the `<tr>` was throwing console warnings about the elements in the list (the `<tbody>` ones) not having keys.

I think this is good now.

PR #308, opened 2023-07-07
Split multiple GAM tracks vertically

### LINKED ISSUE #216: Allow different GAM files to stack vertically, with optional separating line between them within each node

adamnovak:
This changes GAM layout so that multiple tracks are laid out separately in separate passes, one below the other.

It looks like this, where I have an orange GAM and then a purple GAM. There's a 1-read-high margin between/above GAM tracks in each node when multiple GAM tracks are added.

graph(1)

This PR also includes some fixes for deserializing the link to view URLs and for tunring mapping quality coloring on and off, since I ran into related bugs when trying to use the view links to test this.

This will fix #216.

### DIFF
--- a/src/components/CopyLink.js
+++ b/src/components/CopyLink.js
@@ -79,6 +79,6 @@ export const urlParamsToViewTarget = (url) => {
   if (s[1]) {
     result = qs.parse(s[1]);
   }
- 
+
   return result;
 };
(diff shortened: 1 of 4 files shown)


adamnovak:
It looks like adding. separating line would be a lot more work than this; I was able to just run the read layout algorithm per-track for this, whereas to add separating lines we need to actually implement some new D3 stuff I don't know how to do to draw the line.

We will see how it is with just the gutter/margin for now, I think.

Issue #266, opened 2023-05-10
Region field label Z index is too high and it hovers over other elements


adamnovak:
![Image](https://user-images.githubusercontent.com/5062495/237412954-75a70011-f2a0-483e-914c-2624cf6a5db2.png)

If you open the BED file dropdown, for example, the "Region" text from the region picker hovers over it, as seen in this screenshot.

We need to move "Region" lower or move the dropdown contents higher so this doesn't happen.

PR #310, opened 2023-07-10
Bed multiple gam

ducku:
Modified the prepare chunk script to use flags and be able to take multiple gam files. 

Some work on the server side still needs to be done to select tracks specified in the bed file.


adamnovak (changes requested):
The script looks pretty good, but I think the server side changes forgot about how a BED can have multiple lines for different regions. Somewhat relatedly, the refactor of code into the `getChunkPath` function has cut some lines of data flow.

When is the client going to call `getChunkPath`? Right now it looks like the function has a new server endpoint and a client function to call it but the client function isn't used and doesn't seem to explain why it is going to be needed.

scripts/prepare_chunks.sh
```diff
@@ -1,13 +1,56 @@
...
+
+    vg_chunk_params=" $vg_chunk_params -a $GAM_FILE"
+done
+
+
+# Call vg chunk
+vg chunk $vg_chunk_params > $OUTDIR/chunk.vg
+
+# Cosntruct JSON
+JSON_STRING=$(jq -n \
```
adamnovak, line 48: This is really good!

We might need to make sure the user actually has `jq` somehow though. Maybe put it in the documentation for making the chunks that you are going to need `jq` installed? Or if we were really cool the script could test and complain instead of erroring out with a command not found.

ducku:
Still need to change prepare_chunks.sh to detect jq, reconstruct example chunks, and write documentation for taking a collection of small-region graphs and read sets from Lancet.

scripts/prepare_chunks.sh
```diff
@@ -32,23 +32,27 @@ for GAM_FILE in "${GAM_FILES[@]}"; do
     # Put gam files into array format
     if [ -z "$GAM_FILES_JSON" ]
     then
-        GAM_FILES_JSON="$GAM_FILE"
+        GAM_FILES_JSON="[$GAM_FILES_JSON\"$GAM_FILE\""
```
adamnovak, line 35: If it's empty you don't need to include it here.

scripts/prepare_chunks.sh
```diff
@@ -32,23 +32,27 @@ for GAM_FILE in "${GAM_FILES[@]}"; do
...
-        GAM_FILES_JSON="$GAM_FILES_JSON, $GAM_FILE"
+        GAM_FILES_JSON="$GAM_FILES_JSON, \"$GAM_FILE\""
     fi
 
     vg_chunk_params=" $vg_chunk_params -a $GAM_FILE"
 done
 
+GAM_FILES_JSON="$GAM_FILES_JSON]"
+
+echo $GAM_FILES_JSON
```
adamnovak, line 46: It looks like `jq` can convert a file of lines to a JSON array of strings in a way that will properly handle special characters:

```
$ printf 'abc\n123\n",drop table students,"\n' | jq -R '[.]' | jq -n '[inputs[]]'
[
  "abc",
  "123",
  "\",drop table students,\""
]
```

So you might want to try that instead of hand-building JSON in Bash?

You can grab a temporary file with something like `TEMP_FILE_NAME="$(mktemp)"`

adamnovak (changes requested):

src/components/RegionInput.js
```diff
@@ -20,45 +20,63 @@ export const RegionInput = ({
...
+          }
         }}
 
         options={displayRegions}
         renderInput={(params) => (
           <TextField
             {...params}
             label="Region"
-            style={{zIndex:'-100'}}
+            //style={{zIndex:'-100'}}
```
adamnovak, line 75: I think we need this to keep the "Region" label text from drawing over the BED file dropdown when the BED file dropdown is open. I tested your branch here and it seems to reintroduce that bug. Uncommenting this line fixes that but then the region dropdown doesn't pop open when clicked.

If we have to take this out to get the region dropdown to pop open when clicked (because e..g it already is popping open but it is rendering behind the page somehow) then we have to find another way to keep the "Region" label from drawing over the BED file dropdown.

adamnovak (changes requested):
This looks like it reintroduces #266 with the region field changes.

The BED file stuff looks good though. Maybe the region field stuff should move to a new PR and we can merge this one?

adamnovak (approved):

PR #314, opened 2023-07-12
Reorder Header Form

### LINKED ISSUE #303: Reorder header form so it looks nice
The help button seems to have escaped onto its own line, and the track picker button also seems to be getting its own line. We probably want to rearrange things so we can get as many of the buttons as possible onto one line, in an order that makes sense.

shreyasun:
Fixes #303

### DIFF
--- a/src/components/HeaderForm.js
+++ b/src/components/HeaderForm.js
@@ -12,6 +12,7 @@ import RegionInput from "./RegionInput";
 import TrackPicker from "./TrackPicker";
 import SelectionDropdown from "./SelectionDropdown";
 import { parseRegion, stringifyRegion } from "../common.mjs";
+
 // See src/Types.ts
 
 const DATA_SOURCES = config.DATA_SOURCES;
@@ -689,6 +690,7 @@ class HeaderForm extends Component {
               >
                 Data:
               </Label>
+
               <select
                 type="select"
                 value={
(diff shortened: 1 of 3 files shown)


src/components/HeaderForm.js
```diff
@@ -739,6 +736,25 @@ class HeaderForm extends Component {
...
+              {mountedFilesFlag && 
+                <div style={{display: "flex"}}>
+                  <DataPositionFormRow
+                    handleGoLeft={this.handleGoLeft}
+                    handleGoRight={this.handleGoRight}
+                    handleGoButton={this.handleGoButton}
+                    uploadInProgress={this.state.uploadInProgress}
+                    getCurrentViewTarget={this.props.getCurrentViewTarget}
+                    viewTargetHasChange={viewTargetHasChange}
+                  />
```
adamnovak, line 749: I think we should put this in a variable instead of duplicating it.

adamnovak (changes requested):
I think there's a way to avoid duplicating DataPositionFormRow, but otherwise this looks good!

src/components/HeaderForm.js
```diff
@@ -675,6 +675,15 @@ class HeaderForm extends Component {
       this.state.fileSelectOptions
     );
 
+    const DataPositionFormRowComponent = <DataPositionFormRow
```
adamnovak, line 678: I would use `camelCase` for this here I think, like other variables we use in the tube map. But this is probably fine.

adamnovak (approved):

PR #316, opened 2023-07-29
Fix Region Dropdown and Overlapping Issues

ducku:
Fixed Region dropdown onClick by removing its zIndex contraints. 
Fixed bedfile dropdown overlapping with other components by increasing SelectionDropdown's menu zIndex.


src/components/RegionInput.js
```diff
@@ -55,7 +55,7 @@ export const RegionInput = ({
           <TextField
             {...params}
             label="Region"
-            style={{zIndex:'-100'}}
+            //style={{zIndex:'-100'}}
```
adamnovak, line 58:
```suggestion
```

We should just cut the line if we aren't going to use it.

adamnovak (approved):
This looks pretty good. I'm going to cut the dead code and merge it.

PR #324, opened 2023-08-02
Adds client side read coverage statistics

### LINKED ISSUE #218: Client side read coverage statistics
Lancet can annotate graph nodes, and one of the annotations it uses is coverage.

vg can't understand graph node annotations, but it would still be good to be able to see coverage for a node in the tube map. We could do this client-side by counting the displayed reads that visit each node, and then put the information somewhere in the UI (maybe in a tooltip?)

shreyasun:
Fixes #218


src/components/PopUpInfoDialog.js
```diff
@@ -0,0 +1,43 @@
...
+  attributes,
+  close,
+}) => {
+    return(
+      <div>
+        <PopupDialog open={open} close={close}>
+            <h5>Object Information</h5>
+            <table>
+              <tbody>
+                {/* Node info here */}
```
adamnovak, line 16: This is for tracks as well as nodes, right?

src/util/tubemap.js
```diff
@@ -3109,26 +3116,129 @@ function drawNodes(dNodes) {
...
+    let currRead = allReads[readNum];
+    countBases += currRead.finalNodeCoverLength - currRead.firstNodeOffset;
+  }
+  // outgoing reads
+  for (let readVisit of node.outgoingReads) {
+    let readNum = readVisit[0];
+    let currRead = allReads[readNum];
+    // coverage of outgoing read would be the the distance between the end of the node and the
+    //  starting point of the read within the node
+    countBases += node.sequenceLength - currRead.firstNodeOffset;
```
adamnovak, line 3238: I think we're forgetting to skip the read here when this read is also an `incomingReads` read. At this line we add the length of this node minus the offset along the first node. That only makes sense if this *is* the first node on the read, and the read is not also incoming, but we don't seem to check for that.

adamnovak (changes requested):
This looks pretty good, and I like the tests, but I think we're kind of double-counting reads that go straight through a node?

src/util/tubemap.js
```diff
@@ -3109,26 +3116,129 @@ function drawNodes(dNodes) {
...
+  nodeAttributes.push(["Node Length:", currentNode.sequenceLength + " bases"]);
+  nodeAttributes.push(["Haplotypes:", currentNode.degree]);
+  nodeAttributes.push([
+    "Aligned Reads:",
+    currentNode.incomingReads.length +
+      currentNode.internalReads.length +
+      currentNode.outgoingReads.length,
+  ]);
+  nodeAttributes.push([
+    "Number of Reads Visiting Node:",
```
adamnovak, line 3173: This counts reads repeatedly if they visit repeatedly, and should have a name that reflects that.

```suggestion
    "Total Visits:",
```

src/util/tubemap.js
```diff
@@ -3109,26 +3116,129 @@ function drawNodes(dNodes) {
...
+    numReadsVisitNode(currentNode),
+  ]);
+  nodeAttributes.push(["Coverage:", coverage(currentNode, reads)]);
+
+  console.log("Single Click");
+  console.log("node show info callback", config.showInfoCallback);
+  config.showInfoCallback(nodeAttributes);
+}
+
+export function numReadsVisitNode(node) {
```
adamnovak, line 3183:
```suggestion
// Count the number of distinct reads that visit the given node object.
export function numReadsVisitNode(node) {
```

src/util/tubemap.js
```diff
@@ -3109,26 +3116,129 @@ function drawNodes(dNodes) {
...
+  // The first node will not have any incoming reads.
+  for (let readVisit of node.incomingReads) {
+    countReads.add(readVisit[0]);
+  }
+  // internal reads are reads that start and end within the node. They are represented as
+  // an array of values which indicate the read index.
+  for (let read of node.internalReads) {
+    countReads.add(read);
+  }
+  // outgoing reads are reads that exit the node when the read starts within it. They are represented as
```
adamnovak, line 3196: Yeah this is probably the comment we need. I just didn't understand that it was actually true.

src/util/tubemap.js
adamnovak: We don't need to do this. Incoming and outgoing reads are in fact disjoint; incoming reads also includes reads that make it through the node.

src/util/tubemap.test.js
```diff
@@ -211,4 +214,394 @@ describe('cigar_string', () => {
...
+  }
+
+  for (let [readNum, visitIndex] of node.outgoingReads) {
+    if (readNum >= reads.length) {
+      throw new Error("Outgoing read " + readNum + " doesn't exist")
+    }
+    let read = reads[readNum]
+    if (visitIndex >= read.sequenceNew.length) {
+      throw new Error("Outgoing read " + readNum + " visit " + visitIndex + " doesn't exist")
+    }
```
adamnovak, line 278: We should constrain the outgoing reads to start at that node, since that's the semantics actually used.
```suggestion
    }
    if (visitIndex !== 0) {
      throw new Error("Outgoing read did not start at the node it is outgoing for")
    }
```

src/util/tubemap.test.js
```diff
@@ -211,4 +214,394 @@ describe('cigar_string', () => {
...
+    checkNodeName(read, visitIndex)
+    if (visitIndex == read.sequenceNew.length - 1) {
```
adamnovak, line 280:
```suggestion
    if (visitIndex === 0) {
```
I just got this check wrong.

src/util/tubemap.test.js
```diff
@@ -211,4 +214,394 @@ describe('cigar_string', () => {
...
+    expect(coverage(node, reads)).toBe(2.17);
+  })
+  // TEST #4
+  it('can handle node of length of 30 with 2 incoming reads, 4 internal reads, and 3 outgoing reads', async () => {
+    const node = {        
+      nodename: '4',    
+      sequenceLength: 30,
+      incomingReads: [[0, 1], [1, 1]],
+      internalReads: [2, 3, 4, 5],
+      outgoingReads: [[6, 2], [7, 2], [8, 2]],
```
adamnovak, line 455: If outgoing reads are all meant to start in the nodes they are outgoing from, why are these reads listed as outgoing for this node but at visit index 2?

adamnovak (approved):
OK, I am convinced by https://github.com/vgteam/sequenceTubeMap/blob/06657607b92808723cd27d23be068badb6855b24/src/util/tubemap.js#L590-L592 that reads are only ever outgoing from the node they start at, and that we don't actually have a read listed as both incoming and outgoing every time it enters and then leaves a node.

So I think that the coverage algorithm is going to work (though it doesn't account for deletions), and we should be able to merge this after we come up with a set of tests that obey those semantics and also pass.

PR #328, opened 2023-08-15
Highlighted reads

### LINKED ISSUE #323: Draw highlighted reads over other reads
When mousing over a read to highlight it, the read retains its Z position. So other reads can hover over your highlighted read, obscuring it and preventing you from seeing where it goes.

When a read is highlighted, it should draw through or over all the other reads, so you can always see where your highlighted read is going.

shreyasun:
Fixes #323


src/util/tubemap.js
```diff
@@ -511,17 +511,22 @@ function createTubeMap() {
...
-  drawReversalsByColor(trackCorners, trackVerticalRectangles);
-  drawTrackRectangles(trackRectanglesStep3);
-  drawTrackRectangles(trackRectangles, "read");
-  drawTrackCurves("read");
+  // all drawn tracks are grouped 
+  let trackGroup = svg.append("g").attr("class", "track")
+  drawTrackRectangles(trackRectangles, undefined, trackGroup);
+  drawTrackCurves(undefined, trackGroup);
+  drawReversalsByColor(trackCorners, trackVerticalRectangles, undefined, trackGroup);
+  drawTrackRectangles(trackRectanglesStep3, undefined, trackGroup);
```
adamnovak, line 519: I think the `typeof type === "undefined"` checks were meant to check if the argument was omitted or not. So turning around and manually passing `undefined` changes what the code at the other end really means in a needlessly complicated way.

It would make more sense to leave the optional argument last and make the new argument a next-to-last argument, or else to explicitly pass the type here instead of having the argument be optional.

adamnovak (changes requested):
This looks like it should work, but I think passing `undefined` into the various draw functions is excessively complicated.

adamnovak (approved):

Issue #295, opened 2023-06-14
Make BED system able to retrieve pre-made chunks from URLs rather than local folders


adamnovak:
This would make the tube map easier to use with Lancet, because everyone could use one hosted instance of the tube map and as long as they put their Lancet-generated BED files and associated chunk data on the web, we could draw it.

adamnovak:
We can also/as part of this make the track picker take URL strings in addition to files already on the server.

adamnovak:
This might interact kind of poorly with the logic we have right now that looks at what files exist in the chunk directory for e.g. finding all the GAM files. We might need to come up with a scheme where we put an index file in each chunk directory, naming all the other files that are available.

adamnovak:
We also do probably want the BED file dropdown to let you specify a URL *for the BED file itself*, in addition to letting you mount one in.

Relative paths in a BED from a URL should be relative to where the BED file is.

adamnovak:
* We should make sure the user can't use our file fetching to get ahold of web pages that the server can access but the client shouldn't be allowed to see.
* We should make sure the server can't be convinced to download a 10 TB "BED file". So we need to check content-length headers and have some kind of limit.
* We don't want to re-download every file on every request, but we also don't want to let the disk fill up with cached files, and we don't want to be stuck with old data when the user updates a file on their server. So we need caching and also something with `HEAD` or etags or last modified headers to only re-fetch files when they change.

adamnovak:
We want to first do URLs in the BED, and then a URL for the BED, and then letting the user actually provide the URL for the BED in the UI.

Then later we can work on URL tracks more broadly:
* Get rid of data path so URLs and mounted and uploaded files can be mixed
* Have a create track button in the UI that lets you enter a URL (and later upload a file). That adds a new pickable track client-side to the list of available tracks.
* We won't bother trying to persist the list of available created tracks, but we should re-create them when needed for a link-to-view link that uses them.

adamnovak:
Cases to test:

- [ ] A directory of files served with `http-server` from `npm install -g http-server`
- [ ] Something off a server on the Internet over HTTPS, such as in `https://courtyard.gi.ucsc.edu/~anovak` or another location.
- [ ] Something in a Github repo, via e.g. `https://raw.githubusercontent.com/vgteam/sequenceTubeMap/e49eebf3a9b0f2b7c313ec6c6c8e3ebcb031305a/exampleData/chunk-ref-2000-3000/regions.tsv`

[...]

PR #329, opened 2023-08-18
Allow URLs to be inputted as BED files

ducku:
#295 
- URLs can be inputted via pasting in the text field and clicking the create option in the dropdown
- BED files are scanned and discarded
- Chunks specified in the BED file are stored until server termination


src/server.mjs
```diff
@@ -1154,13 +1303,27 @@ function getBedRegions(bedFile, dataPath) {
     }
     bed_info["chunk"].push(chunk);
 
+  });
 
+  // download chunks if bedfile is an url
+  if (isURL) {
+    await retrieveChunks(bedFile, bed_info["chunk"]);
```
adamnovak, line 1310: Does it make sense to retrieve all the BED chunk data up front? We currently seem to limit the BED file size to a bit over 4 kilobytes, but really we're going to want to support cases like a BED file of 20,000 different genes. So as soon as we raise the BED file size limit we're going to want to come up with a way to lazily fetch just the chunks people actually go to.

src/server.mjs
```diff
@@ -1154,13 +1303,27 @@ function getBedRegions(bedFile, dataPath) {
...
+  // check for a tracks.json file to prefill tracks configuration
+  for (let i = 0; i < bed_info["chunk"].length; i++) {
     let tracks = {};
     let trackID = 1;
 
-    const chunk_path = `${dataPath}${chunk}`;
-    const track_json = path.join(chunk_path, "tracks.json");
+    const dirName = bed_info["chunk"][i];
+    const chunk_path = `${dataPath}${dirName}`;
+    let track_json = `${chunk_path}/tracks.json`
```
adamnovak, line 1320: I think this is the third place we manually paste together stuff from the BED line to get the path to a file. The third time you just implement the same thing is about when it's time to make it a function. We probably want a function we can call to get us the local path for a file, to either put it at or read it from.

src/server.mjs
```diff
@@ -1094,17 +1109,139 @@ api.post("/getPathNames", (req, res, next) => {
...
+
+    // retrieve and download each file in the chunk directory
+
+    let chunkContentURL = dataPath + "/" + dirName + "/chunk_contents.txt";
+    let response = await fetch(chunkContentURL, options);
+
+    const chunkContent = await response.text();
+    const fileNames = chunkContent.split("\n");
+
+    for (const fileName of fileNames) {
```
adamnovak, line 1183: We probably want some limit on the number of file names here.

src/server.mjs
```diff
@@ -1094,17 +1109,139 @@ api.post("/getPathNames", (req, res, next) => {
...
+    };
+
+
+
+    // retrieve and download each file in the chunk directory
+
+    let chunkContentURL = dataPath + "/" + dirName + "/chunk_contents.txt";
+    let response = await fetch(chunkContentURL, options);
+
+    const chunkContent = await response.text();
```
adamnovak, line 1180: It looks like there's no limit on the size of the chunk contents file that we will accept. It could be 30 gigabytes and we will run out of memory to store it and crash.

src/server.mjs
```diff
@@ -1231,6 +1393,9 @@ export function start() {
       connections: undefined,
       // Shut down the server
       close: async () => {
+        // remove the temporary directory
+        fs.rmSync(config.tempDirPath, { recursive: true, force: true });
```
adamnovak, line 1397: Cleaning the whole thing at the end is a pretty good idea.

scripts/prepare_chunks.sh
```diff
@@ -54,4 +54,10 @@ JSON_STRING=$(jq -n \
...
 
-printf "%s\n" "$JSON_STRING" > $OUTDIR/tracks.json
\ No newline at end of file
+printf "%s\n" "$JSON_STRING" > $OUTDIR/tracks.json
+
+
+for file in `ls $OUTDIR/`
+do
+    printf "$file\n" >> $OUTDIR/chunk_contents.txt
+done
```
adamnovak, line 63: Do we have our own loop just to stop `chunk_contents.txt` from listing itself?

adamnovak (changes requested):
This looks pretty good, but I think the file management/sandboxing isn't quite robust enough.

1. Because there's no unified safe fetch wrapper function, we only limit content length on the BED file itself, and we don't seem to do things like check for unsuccessful response codes.
2. The path manipulation is done with a mixture of `+` and `path.resolve`, neither of which will properly protect against attacks like using `../` or a leading `/` to place stuff outside of the dedicated sandbox directories for the downloads. We should have a function that gets a target path based on like a chunk-local filename, some identifier for the chunk, and some identifier for the BED file, and guarantees that it will be actually inside the directory we want it in, and we should consistently use that.
3. The same chunk name seems like it will collide if it is used in different BED files at the same time. We might want to use some hashing to make up unique directories per BED.
4. I don't think downloading all the chunks up front (and in serial!) is going to scale to larger BED files. We probably want to download chunks when needed, and download all the files in the chunk in parallel (by launching the downloads and then awaiting all of them).

src/server.mjs
```diff
@@ -666,13 +670,32 @@ function getChunkPath(bedFile, dataPath, parsedRegion) {
...
         break;
       }
     }
   }
+
+  // if chunk is pulled from an url
+  const chunkDirPath = path.resolve(config.tempDirPath, chunkPath);
+  const baseURLPath = path.resolve(chunkDirPath, "baseURL.txt");
+  // if the chunk directory exists and the baseURL.txt is in it
+  if (fs.existsSync(chunkDirPath) && fs.existsSync(baseURLPath)) {
```
adamnovak, line 685: I don't think the `baseURLPath` file can exist without the parent directory existing.

src/server.mjs
```diff
@@ -654,9 +657,10 @@ function returnErrorMiddleware(err, req, res, next) {
 app.use(returnErrorMiddleware);
 
 // Gets the chunk path from a region specified in a bedfile
```
adamnovak, line 659: This comment makes this look like it is a pure function that determines what path on the local disk some data for a BED file ought to live at.

Actually it is responsible for downloading that data if it doesn't exist. That's surprising and we should probably include in the comment something about how this function ensures that the chunk data is available on the local disk and returns a path to it.

src/server.mjs
```diff
@@ -1154,13 +1322,31 @@ function getBedRegions(bedFile, dataPath) {
...
+      // 1 level above the bedurl
+      const baseURL = bedFile.split('/').slice(0, -1).join('/') + '/';
+      await retrieveChunk(baseURL, dirName, false);
+    }
+
+    const tempTracksPath = path.resolve(config.tempDirPath, sanitize(dirName), "tracks.json");
+    // check the temporary directory for downloaded tracks info
+    if (fs.existsSync(tempTracksPath)) {
+      track_json = tempTracksPath;
+    }
```
adamnovak, line 1348: What if this file happens to exist, but the BED file didn't come from a URL and was actually a local file?

src/server.mjs
```diff
@@ -1094,17 +1118,148 @@ api.post("/getPathNames", (req, res, next) => {
...
+    dataRead.push(value);
+    bytesRead += new Blob([value]).size;
+
+    if (bytesRead > maxBytes) {
+      reader.cancel();
+      throw new Error("Fetched request failed, maximum file size exceeded: " + maxSize);
+    }
+  }
+
+  return new Response(new Blob(dataRead), { headers: response.headers });
```
adamnovak, line 1199: Maybe the interface of this function should change to just return a Blob?

Alternately, if we want to support downloading files that are larger than is convenient to hold in memory, but still use a size limit, we'd want to return a stream from this function, but with a size-limiting filter on the stream that will raise an error when the reader goes to read past the size limit.

src/server.mjs
```diff
@@ -1094,17 +1118,148 @@ api.post("/getPathNames", (req, res, next) => {
...
+
+}
+
+// downloads files fo the specified chunk name from the destination of the bedURL
+// includeContent only downloads the tracks.json file when set to false
```
adamnovak, line 1204: I don't think this is a good design. When used with a false flag, this isn't *really* retrieving the chunk in a meaningful sense.

If we actually do want to create N directories, each with a base URL file and a track metadata file, when we first see a BED file with N entries, then we should split that out from downloading the actual chunk files. We don't need the code to make the base URL file when actually retrieving the full chunk, and we don't need the code to look at the `chunk_contents.txt` file when retrieving `tracks.json` (because we already know the name and we can just fetch it if it exists).

src/server.mjs
```diff
@@ -1094,17 +1118,148 @@ api.post("/getPathNames", (req, res, next) => {
...
+  // path to the designated chunk in the temp directory
+  const chunkDir = path.resolve(config.tempDirPath, sanitize(chunk));
+
+  // TODO: check if this chunk has been downloaded before, to prevent duplicate fetches
+  if (!fs.existsSync(chunkDir)) {
+    fs.mkdirSync(chunkDir, { recursive: true });
+  } 
+
+  // baseURL/dirName/"chunk_contents.txt"
+  let chunkContentURL = new URL(path.join(sanitize(chunk), "chunk_contents.txt"), baseURL).toString();
```
adamnovak, line 1216: I don't think the sanitize is appropriate here when creating the URL to fetch stuff from. I think `/` and `..` are meant to be allowed in the URL that the BED file points to for its chunk. This looks like it is going to strop them out.

We also want to support e.g. `baseURL/dirName/subDirName/chunk_contents.txt`.

src/server.mjs
```diff
@@ -1094,17 +1118,148 @@ api.post("/getPathNames", (req, res, next) => {
...
+    }
+  }
+
+  // write the baseURL to a file
+  const baseURLFile = path.resolve(chunkDir, "baseURL.txt");
+  fs.writeFile(baseURLFile, baseURL, function (error) {
+    if (error){
+      throw error;
+    }
+  })
```
adamnovak, line 1241: I don't understand why this system is necessary.

When the user asks to get chunk data for a tube map rendering, they send along the BED URL, and we are already finding the right region in the BED file to get the chunk information so that we know where to look on the server for the downloaded chunk data. So we can always make the base URL of the chunk with a `new URL()` based on the BED file's own URL and whatever it lists as the path to the chunk, right? Why do we need to make it once and save it in a flat file for every chunk?

src/server.mjs
adamnovak: I feel like `getChunkPath` *wants* to be the function that, given a BED URL and a chunk path string from that BED, determines the local path at which to look for stuff. But we can't use it here because it also does other stuff like actually downloading the files. Maybe it needs to get split up?

adamnovak:
Plan:
* Make sure you can't "go" without any tracks
* Make having tracks.json mandatory for a premade chunk
* Blank out the tracks when the user picks a BED region, and request the tracks for that chunk from the server. (We might need to tall the client if the BED regions have/don't have premade chunks and only do this for the ones that do)
* When the tracks for the chunk arrive, put them in, which will re-enable "go"

Then we will be able to only fetch tracks for the BED regions we actually need them for.

adamnovak:
We talked about this more and now the plan is to merge this as soon as the minor problems/cross-talk issues are fixed, and *then* refactor so we can download large files not to memory/avoid downloading all the track data up front, since neither of those seems like it will be a problem for Lancet data in particular.

adamnovak:
OK, I've refactored this a bit to convince myself that we won't end up letting people write stuff on the server outside of where they are supposed to. I also fixed error handling, which doesn't magically work for async API functions we pass to Express since we don't have Express 5 yet (it's still in beta). I think eventually we will have to move all the guts of the tube map from `next()`-based error handling and careful callback ordering to promises and async/await, but I am not going to do that today.

Also I think I found a use case for premade chunks with no track JSON, which is that `cactus.bed` seems to use one, which got deleted and which I am restoring.

PR #332, opened 2023-08-31
Highlight reference paths

### LINKED ISSUE #217: Visually highlight "reference" paths to distinguish them from e.g. tumor or normal sample paths
If an incoming GFA tags some paths as "reference" (and thus the vg file has a reference sense for them in its internal path metadata) we want to be able to highlight them differently on the tube map, vs. e.g. other generic or haplotype-sense paths still indexed for random access.

shreyasun:
Fixes #217


adamnovak (approved):
This looks good!

PR #335, opened 2023-09-11
Color palette preferences presets

### LINKED ISSUE #235: Preferences or preset support for color palettes
The tube map has support for a JSON config for the tracks to be displayed.

Lancet would benefit from being able to send track color settings along with this, so that Lancet could compose a view of some GAMs and a graph, and then the tube map could display the view of the GAM reads with appropriate tumor/normal coloration.

This could also be a place to hide node color override information.

shreyasun:
Fixes #235

### DIFF
--- a/scripts/prepare_chunks.sh
+++ b/scripts/prepare_chunks.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-
 while getopts x:h:g:r:o: flag
 do
     case "${flag}" in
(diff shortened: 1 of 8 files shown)


scripts/prepare_chunks.sh
```diff
@@ -29,34 +28,25 @@ mkdir -p $OUTDIR
 
 vg_chunk_params="-x $XG_FILE -g -c 20 -p $REGION -T -b $OUTDIR/chunk -E $OUTDIR/regions.tsv"
 
-GAM_FILES_STRING=""
+# construct track JSON for xg file
+jq -n --arg trackFile "${XG_FILE}" --arg trackType "graph" --argjson trackColorSettings '{"mainPalette": "greys", "auxPalette": "ygreys"}' '$ARGS.named' >> $OUTDIR/tracks.json
+
+# construct track JSON for gbwt file; if not any specific gbwt file, then default would be haplotype
+if [[ ! -z "${GBWT}" ]] ; then
+    jq -n --arg trackFile "${GBWT}" --arg trackType "haplotype" --argjson trackColorSettings '{"mainPalette": "blues", "auxPalette": "reds"}' '$ARGS.named' >> $OUTDIR/tracks.json
```
adamnovak, line 36: Maybe these should be rainbow colors instead?

src/server.mjs
```diff
@@ -1425,38 +1426,20 @@ async function getBedRegions(bed) {
...
-          tracks[trackID]["trackType"] = fileTypes["READ"];
-          tracks[trackID]["trackColorSettings"] = {...config.defaultReadColorPalette};
-          trackID += 1;
-        }
-
+        // Create string of tracks data
+        const string_data = fs.readFileSync(track_json);
+        const parser = new JSONParser();
+        parser.onValue = (value, key, parent, stack) => {
+          if (stack > 0) return; // ignore inner values
```
adamnovak, line 1437: I might format this with braces, more like the rest of the code instead of the example it originally came from.

adamnovak (approved):
I think this looks OK. We just have to fix the problem with Jest not wanting to work with the parser library.

PR #336, opened 2023-09-14
Node coverage stats deletions

### LINKED ISSUE #327: Account for deletions in per-node coverage
In https://github.com/vgteam/sequenceTubeMap/pull/324 we don't account for bases of a node being deleted in a read.

We should count the deleted bases and make sure that they aren't counted towards coverage on the node. If one read visits a node and half of the node sequence is deleted in the read, coverage should be 0.5.

shreyasun:
Fixes #327


adamnovak (approved):
The code to deduct for deletions could be refactored out into one function instead of 3 copies, but I think this is fine and going to work.

PR #339, opened 2023-09-25
Issues/247 Remove Seperation of Datapath and Filename in Requests

ducku:


src/components/HeaderForm.js
```diff
@@ -310,20 +307,20 @@ class HeaderForm extends Component {
       } else {
         json.bedFiles.unshift("none");
 
-        if (this.state.dataPath === "mounted") {
+        if (this.state.dataType === dataTypes.MOUNTED_FILES) {
```
adamnovak, line 310: Should we rename `MOUNTED_FILES` to maybe `CUSTOM`?

adamnovak (changes requested):
This looks pretty good.

It might make sense to rebase f563dabdb25d2d16e3fec03091bf5cbb0204e72f out and save it for the PR that actually unifies uploaded and mounted data in the UI.

PR #342, opened 2023-09-27
Color nodes

### LINKED ISSUE #234: Color nodes
We would like to be able to color nodes in the tube map display, for example to allow tumor-private nodes to stand out, to enable quickly finding them from an overview view.

This could be pre-generated by Lancet as custom GFA tags.

This would require new machinery in vg to propagate the annotations through chunking or hide them somewhere when they can't live in a .vg or .xg file.

shreyasun:
Fixes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ This will save some time during the interactive visualization, especially if the
 
 The net result needs to be one or more chunk directories on disk, referenced from a BED file.
 
-To generate each chunk, you can use the `prepare_chunks.sh` script. You ought to run it from the directory containing your input files and where your output chunks will be stored (i.e. the `dataPath` in `sequenceTubeMpas/src/config.json`), which defaults to the `exampleData` directory in the repo.
+To generate each chunk, you can use the `prepare_chunks.sh` script. You ought to run it from the directory containing your input files and where your output chunks will be stored (i.e. the `dataPath` in `sequenceTubeMaps/src/config.json`), which defaults to the `exampleData` directory in the repo.
 
 For example:
 
@@ -161,6 +161,23 @@ chr1	101	200	Region B	chunk-chr2-101-200
 Note each column is seperated by tabs
 
 This BED file needs to be in the `dataPath` directory, or it can be hosted on the web along with its chunk directories and accessed via URL.
+
+If you want certain nodes of the graph to be colored, place the node names to be colored in a `nodeColors.tsv` file, with a node name on each line, within output directory of the chunk. When rendered, these specified nodes will be colored differently than other nodes.
+
+You can use `prepare_chunks.sh` script to generate this additional `nodeColors.tsv` by adding an additional option. Here is an example:
+
+```
+cd exampleData/
+../scripts/prepare_chunk.sh -x mygraph.xg -h mygraph.gbwt -r chr1:1-100 -d 'Region A' -o chunk-chr1-1-100 -g mygam1.gam -g mygam2.gam -n "1 2 3" >> mychunks.bed
+```
+
+Adding this additional `n` flag will allow a string space delimited input of node names which will be outputted to `nodeColors.tsv`.
+
+```
+1
+2
+3
+```
 
 ##### Pre-made subgraphs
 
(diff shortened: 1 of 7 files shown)


scripts/prepare_chunks.sh
```diff
@@ -88,4 +99,6 @@ done
 
 # Print BED line
 cat $OUTDIR/regions.tsv | cut -f1-3 | tr -d "\n"
+cat $OUTDIR/nodeColors.tsv
+
```
adamnovak, line 103: I don't think we want these inserted here; we won't get the right BED line out on standard output.

Also the file might not exist.

```suggestion
```

README.md
```diff
@@ -139,7 +139,7 @@ This will save some time during the interactive visualization, especially if the
 
 The net result needs to be one or more chunk directories on disk, referenced from a BED file.
 
-To generate each chunk, you can use the `prepare_chunks.sh` script. You ought to run it from the directory containing your input files and where your output chunks will be stored (i.e. the `dataPath` in `sequenceTubeMpas/src/config.json`), which defaults to the `exampleData` directory in the repo.
+To generate each chunk, you can use the `prepare_chunks.sh` script. You ought to run it from the directory containing your input files and where your output chunks will be stored (i.e. the `dataPath` in `sequenceTubeMaps/src/config.json`), which defaults to the `exampleData` directory in the repo.
```
adamnovak, line 142: Good catch!

src/server.mjs
```diff
@@ -1001,6 +1038,8 @@ function cleanUpAndSendResult(req, res, next) {
     result.graph = req.graph;
     result.gam = req.withGam === true ? req.gamResults : [];
     result.region = req.region;
+    result.coloredNodes = req.coloredNodes;
+    // add new nodenames field
```
adamnovak, line 1042:
```suggestion
```

I think the line this comment describes isn't here? Or is it meant to be about the previous line?

src/util/tubemap.js
```diff
@@ -3145,14 +3153,28 @@ function drawNodes(dNodes, groupNode) {
...
-    .style("stroke", "black")
+    .style("fill", (d) => colorNodes(d.name)["fill"])
+    .style("fill-opacity", (d) => colorNodes(d.name)["fill-opacity"])
+    .style("stroke", (d) => colorNodes(d.name)["outline"])
     .style("stroke-width", "2px")
     .append("svg:title")
     .text((d) => getPopUpNodeText(d));
 }
 
+function colorNodes(nodeName){
```
adamnovak, line 3164: We should explain the new function.

```suggestion
// Given a node name, return an object with "fill", "fill-opacity", and "outline"
// keys describing what colors should be used to draw it.
function colorNodes(nodeName){
```

src/util/tubemap.js
```diff
@@ -3145,14 +3153,28 @@ function drawNodes(dNodes, groupNode) {
     .on("mouseout", nodeMouseOut)
     .on("dblclick", nodeDoubleClick)
     .on("click", nodeSingleClick)
-    .style("fill", config.transparentNodesFlag ? "none" : "#fff")
-    .style("fill-opacity", config.showExonsFlag ? "0.4" : "0.6")
-    .style("stroke", "black")
+    .style("fill", (d) => colorNodes(d.name)["fill"])
+    .style("fill-opacity", (d) => colorNodes(d.name)["fill-opacity"])
+    .style("stroke", (d) => colorNodes(d.name)["outline"])
```
adamnovak, line 3158: I'm not sure I see a great way around this, but this is going to call the coloring function three times for every node, each time getting every color attribute, and then each time throw away the two irrelevant color attributes and just take the one that is used.

We should think about if there's a way to redesign things to avoid this. Maybe we *should* put the color information actually inside the node objects, because that would simplify this.

adamnovak (approved):
This looks pretty good!

I'm again assuming you tested this manually @shreyasun. It might be good to have some kind of automated test, though I'm not sure how we would ask what color something is in D3 in the tests.

I'm going to apply some minor suggestions and merge.

scripts/prepare_chunks.sh
adamnovak, line 103:
```suggestion
```

src/util/tubemap.js
adamnovak: I think using `styles()` might let us make the call once and apply a whole object.

Issue #338, opened 2023-09-20
We should have a form or drop target to upload files, and uploaded files should just appear in your list of files to pick alongside the mounted files.


adamnovak:
We should have a form or drop target to upload files, and uploaded files should just appear in your list of files to pick alongside the mounted files.

_Originally posted by @adamnovak in https://github.com/vgteam/sequenceTubeMap/issues/247#issuecomment-1601152955_

adamnovak:
This is the front-end side of #247.

PR #343, opened 2023-10-02
Issues/338 upload track files

ducku:
#338 

- Remove the current custom upload page
- Allow files to be uploaded through the custom track system


src/components/PickerTypeDropdown.js
```diff
@@ -0,0 +1,54 @@
+import PropTypes from "prop-types";
+import Select from "react-select";
+import React from "react";
+
+export const PickerTypeDropdown = ({
```
adamnovak, line 5: This probably needs a doc comment.

Also, are you sure we really need a new component for this? It looks like this is just a dropdown with two options that gets used once. Why isn't it the same component as e.g. the file type selection dropdown?

src/components/TrackList.js
```diff
@@ -58,6 +66,10 @@ TrackList.propTypes = {
   availableColors: PropTypes.array,
   onChange: PropTypes.func.isRequired,
   onDelete: PropTypes.func.isRequired,
+  showFileSizeAlert: PropTypes.func.isRequired,
+  setUploadInProgress: PropTypes.func.isRequired,
+  getPathNames: PropTypes.func.isRequired,
+  apiUrl: PropTypes.string.isRequired,
```
adamnovak, line 72: And we're also punching all these props through several layers of components to get them to where they are used.

src/config.json
```diff
@@ -73,6 +73,15 @@
...
+  },
+
+  "fileTypeToExtensions": {
+    "graph": ".xg,.vg,.hg,.gbz,.pg",
+    "haplotype": ".gbwt,.gbz",
+    "read": ".gam"
+  },
+
+  "MAXUPLOADSIZE": 5242880,
+  "pickerTypeOptions": ["mounted", "upload"]
```
adamnovak, line 85: What is this for? Is it meant to let you turn off the upload feature? Because it won't on the server side.

adamnovak (changes requested):
This looks pretty good, but I think instead of bringing everything we need to talk to the API down into the track picker control, we should just bring a function to upload files, and leave all the API client stuff and status/error stuff at a higher level in the component hierarchy.

src/components/TrackFilePicker.demo.js
```diff
@@ -15,7 +15,7 @@ export default (<Demo
 
     value: P.string("fileA1.vg"),
     fileType: P.choices(["graph", "haplotype", "read"]),
-    pickerType: P.choices(["dropdown", "upload"]),
+    pickerType: P.choices(["mounted", "upload"]),
```
adamnovak, line 18: Looks like e.g. `showFileSizeAlert` is now required, but isn't sent by the demo. And there isn't really a good value to use for it.

src/end-to-end.test.js
```diff
@@ -390,3 +399,90 @@ it("can retrieve the list of mounted graph files", async () => {
...
+  });
+
+  // select the upload option
+  await waitFor(() => screen.getByText("upload"));
+  fireEvent.click(screen.getByText("upload"));
+
+  await waitFor(() => screen.queryByTestId("file-select-component1"));
+
+  const fileUploader = screen.queryByTestId("file-select-component1");
+  const fakeFile = new File(["example_data"], "example.vg", { type: "text/vg" });
```
adamnovak, line 443: I don't think `text/vg` is right; it should really be `application/octet-stream`, which is probably what the browser is probably going to see it as.

src/end-to-end.test.js
```diff
@@ -390,3 +399,90 @@ it("can retrieve the list of mounted graph files", async () => {
...
+
+  // try to fetch and upload a vg graph file
+  const file_url = "https://raw.githubusercontent.com/vgteam/sequenceTubeMap/9f9f9e8724329d3c7cae9d724e6f8235b275f8d3/exampleData/cactus.vg";
+  const file = await getFileFromURL(file_url, "cactus.vg");
```
adamnovak, line 447: I don't think we want the test to download the data file from the Internet to get the bytes to then upload to the test server. We should be reading the bytes off the local disk instead.

src/end-to-end.test.js
```diff
@@ -390,3 +399,90 @@ it("can retrieve the list of mounted graph files", async () => {
...
+
+  await waitFor(() => {
+    userEvent.upload(fileUploader, file);
+  });
+  
+  // make sure the file is in the upload component
+  expect(fileUploader.files.length).toBe(1);
+  expect(fileUploader.files[0]).toStrictEqual(fakeFile);
```
adamnovak, line 455: Why does this succeed? Don't we upload `file`, which is a different thing here than `fakeFile`?

adamnovak (changes requested):
It looks like the file upload handling logic is in the right place now, and the props are cleaner.

But it looks like we manage to test file upload by *downloading a file from the Internet*, and then taking *that* and uploading it to the test server. That's not right, we really should be opening the file from the local disk instead.

It could be that that is the source of the too-slow-for-CI test that is failing.

adamnovak (approved):

PR #346, opened 2023-10-07
Different tracks colors

### LINKED ISSUE #333: Use different default colors for different track types
In the solution for https://github.com/vgteam/sequenceTubeMap/issues/217 we found that, if you make a new track and then set its type, it keeps whatever default colors there are for a default created track. It doesn't magically adopt the default colors appropriate for the type of track you have made it be.

We should allow the default colors to differ depending on track type, and if you change track types it should reset the colors to sensible defaults for that track type.

shreyasun:
Fixes #333


adamnovak (approved):
This looks good, except the new default colors have changed the correct results of some of the tests, so the tests are failing.

@shreyasun Can you run the tests and fix the right answers so they pass?

adamnovak:
I'm working on diagnosing the tests, and after fixing the missing import, I'm now running into trouble where the file upload test isn't actually managing to do the file upload. I'm not sure why it has decided to fail for this PR, though, since it seems unrelated. I'm going to see if I can fix it.

Issue #352, opened 2023-10-11
Implement cache expiration for downloaded files


adamnovak:
When we download files from a URL, right now we just keep them there forever, until the server restarts.

We should implement some way to expire and delete old downloaded files. It might require some kind of periodic timer in the server.

We might also want a way to limit the total size of the downloaded files, and evict old files when downloading new ones that might put us over the limit.

adamnovak:
We want to add a reader-writer type lock to protect the cache, so we don't delete files that we are about to read to handle requests.

The deletion scan is the writer, and it locks the cache while it scans and deletes old files/folders.

The request-handling functions are the readers, and they lock the cache whenever they start dealing with files that are or might be in the cache, and unlock it afterward.

We probably need to find a Node module that gives us a good reader-writer lock implementation that plays well with holding the lock across callback chains.

We already know adding stuff to the cache is safe under concurrency because we write the files as single operations without yielding during the write.

adamnovak:
@ducku For a usable-looking readers-writer lock implementation, there is https://www.npmjs.com/package/readers-writer-lock

You can do things like:
```
console.log("Getting the write lock...")
await the_lock.write(async function() {
    console.log("I now have the write lock and can do async things while holding it")
})
console.log("I have released the lock")
```

To use it, you need to have all the stuff you want to do while holding the lock be an async function. Right now a lot of the processing of the downloaded chunks happens across a few functions that call each other in sequence from callbacks. We'd have to put a pretty reliable wrapper around all that so that we could have a promise that resolves at the end and rejects if *any* error is encountered. Or, we'd need to do the refactoring I was talking about at https://github.com/vgteam/sequenceTubeMap/issues/196#issuecomment-1755790971, to actually structure the different passes of handling the different files as async functions.

PR #358, opened 2023-10-23
add timed check for expired files

ducku:
#352 

- Deletion of files past a certain expiration time
- Read write locks for cached files


src/server.mjs
```diff
@@ -113,6 +114,36 @@ var limits = {
 };
 var upload = multer({ storage, limits });
 
+// runs every hour
+// deletes any files in the download directory past the set fileExpirationTime set in config
+cron.schedule('0 * * * *', () => {
+  console.log("cron scheduled check");
+  const currentTime = new Date().getTime();
+  // loop through these specified directories
+  for (const dir of [DOWNLOAD_DATA_PATH, UPLOAD_DATA_PATH]) {
```
adamnovak, line 123: Should we maybe warn people near where these are configured that the tube map server needs to use them exclusively, and will delete anything else in there?

src/server.mjs
```diff
@@ -113,6 +114,36 @@ var limits = {
...
+    fs.readdir(dir, (err, files) => {
```
adamnovak, line 124: This only handles one flat directory, right? Don't we download files to different subdirectories?

We have an `mkdirSync()` here:
https://github.com/vgteam/sequenceTubeMap/blob/0280401b5ab6da396231a06a2c9b3f6d6596418b/src/server.mjs#L1285-L1288

But this cleanup code has nothing to loop through the subdirectories recursively, and no `rmdir` to remove the emptied directories.

src/server.mjs
```diff
@@ -113,6 +114,36 @@ var limits = {
...
+      if (!files || files.length === 0) {
+        return;
+      }
+
+      files.forEach((file) => {
+        const filePath = path.join(dir, file)
+        // get file statistics
+        fs.stat(filePath, (statErr, stats) => {
+          const creationTime = stats.birthtime.getTime();
+          if (currentTime - creationTime >= config.fileExpirationTime) {
```
adamnovak, line 135: When we properly support subdirectories, a parent directory might be expired before all its contents are, because the parent directory was created before the files in it were created. But it's not possible to delete the parent directory while there are things inside it. So we'd need to handle that somehow.

We might want to handle expiration at the chunk directory/individual file upload level, and remove whole chunk directories as a unit.

adamnovak (changes requested):
This looks like a good start, but I don't think it handles how chunks are downloaded into directories, and not just all dumped into files in the root of the `DOWNLOAD_DATA_PATH`.

We also don't have anything that handles the case where something is expiring at the same time as it is being used. The chunk download code assumes that if it sees the chunk files downloaded at the beginning of the request, they will still be there when it goes to open them and will not have been removed in the intervening time. If we have a delete job that can run in the background, that won't necessarily be true anymore.

We might need a way to mark a chunk as being used by a request (maybe with `touch` and an assumption that no request can last longer than the expiration time?) so it isn't deleted during a request that uses it, or we need to make accesses to chunk data handle the case where the file goes away and go get it again (which is harder).

adamnovak (changes requested):
There's a missing `await` in the `api.post("/getChunkedData")` `lockDirectories` callback that is going to make the lock be released before the request has been fully handled.

Other than that, it looks like this should work. I've pointed out a couple places where we don't explicitly state guarantees that we later rely on to avoid deadlocks or to keep critical sections protected by the locks. It might make sense to document those sections better.

PR #359, opened 2023-11-03
prevent repeated fetch of downloaded files

ducku:
- Storing of ETags from fetch requests, preventing fetching files that have not been updated and are already on disk

### DIFF
--- a/src/server.mjs
+++ b/src/server.mjs
@@ -4,6 +4,8 @@
 
 "use strict";
 
+import "./config-server.mjs";
+import { config } from "./config-global.mjs";
 import assert from "assert";
 import { spawn } from "child_process";
 import express from "express";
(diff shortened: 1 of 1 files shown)


src/server.mjs
```diff
@@ -1432,21 +1459,9 @@ async function getBedRegions(bed) {
...
-            throw new BadRequestError('Non-track object in tracks.json: ' + JSON.stringify(value))
-          }
-          // put tracks in array
-          tracks_array.push(value);
-        };
-        parser.write(string_data);
+
         // Convert to object container like the client component prop types expect
-        tracks = {...tracks_array}; 
+        tracks = JSON.parse(string_data);
```
adamnovak, line 1464: Is this meant to be part of this PR?

adamnovak (changes requested):
The ETag stuff looks great!

But there are also some changes to track JSON file parsing which it is not clear are intended to be included. Are they meant to be there?

adamnovak:
@ducku said that all the changes are meant to be in this.

PR #366, opened 2023-11-13
Lazy Bed Regions Download

### LINKED ISSUE #353: Make BED region download lazy to alow thousands of premade BED regions
Right now, when we download a BED file from a URL, we immediately try to download some metadata about each region.

We need to make this lazy instead, so that we only try to download anything about a region when it is actually accessed.

This will be required to support BED files with thousands of regions, which the Lancet team needs.

ducku:
- Chunk metadata from the internet is no longer downloaded until region selection
- GetChunkedData now confirms tracks from a bedURL as user might've pressed Go before tracks are fetched
- Closes #353 

Summary of the current process for bedURLs:
- Upon receiving a bedURL, only the bed file is fetched and bedinfo such as regions are returned 
- Upon selection of a bed region, the respective tracks.json file for that chunk is fetched and tracks are loaded into the track picker if applicable
- Upon receiving a request to getChunkedData, if a bedURL is present, the tracks.json file is re-fetched(if necessary) and used instead of the given tracks from the request in case the user hit the GO button before tracks were loaded into the track picker
- getChunkedPath is called when a bed file is included in the request, downloading the remaining chunk if the bed is a URL and returning a chunk path pointing to the location of the download

### DIFF
--- a/src/components/HeaderForm.js
+++ b/src/components/HeaderForm.js
@@ -10,7 +10,7 @@ import ExampleSelectButtons from "./ExampleSelectButtons";
 import RegionInput from "./RegionInput";
 import TrackPicker from "./TrackPicker";
 import BedFileDropdown from "./BedFileDropdown";
-import { parseRegion, stringifyRegion, isValidURL } from "../common.mjs";
+import { parseRegion, stringifyRegion } from "../common.mjs";
 
 
 // See src/Types.ts
@@ -513,7 +513,7 @@ class HeaderForm extends Component {
   // In addition to a new region value, also takes tracks and chunk associated with the region
   // Update current track if the new tracks are valid
   // Otherwise check if the current bed file is a url, and if tracks can be fetched from said url
-  // Tracks remain unchanged if neither conditions are met
+  // Tracks remain unchanged if neither condition is met
   handleRegionChange = async (value, tracks, chunk) => {
     // After user selects a region name or coordinates,
     // update path, region, and associated tracks(if applicable)
@@ -532,20 +532,27 @@ class HeaderForm extends Component {
     // Override current tracks with new tracks from chunk dir
     if (tracks) {
       this.setState({ tracks: tracks });
-      console.log("New tracks have been population");
-    } else if (isValidURL(this.state.bedFile)){
-      // Try to retrieve tracks if the current bed file is an URL
+      console.log("New tracks have been applied");
+    } else {
+      // Try to retrieve tracks from the server
       const json = await fetchAndParse(`${this.props.apiUrl}/getChunkTracks`, {
         method: "POST",
         headers: {
           "Content-Type": "application/json",
         },
-        body: JSON.stringify({ bedURL: this.state.bedFile, chunk: chunk }),
+        body: JSON.stringify({ bedFile: this.state.bedFile, chunk: chunk }),
       });
 
       // Replace tracks if request returns non-falsey value
       if (json.tracks) {
-        this.setState({ tracks: json.tracks });
+        this.setState((laterState) => {
+          if (laterState.region === coords) {
+            // The user still has the same region selected, so apply the tracks we now have
+            return { tracks: json.tracks };
+          }
+          // Otherwise, don't apply the downloaded tracks, because they are no longer relevant.
+          // TODO: Save the downloaded tracks in case the user selects the region again?
+        });
       }
     }
   };
--- a/src/components/HeaderForm.js
+++ b/src/components/HeaderForm.js
@@ -10,7 +10,7 @@ import ExampleSelectButtons from "./ExampleSelectButtons";
 import RegionInput from "./RegionInput";
 import TrackPicker from "./TrackPicker";
 import BedFileDropdown from "./BedFileDropdown";
-import { parseRegion, stringifyRegion } from "../common.mjs";
+import { parseRegion, stringifyRegion, isEmpty } from "../common.mjs";
 
 
 // See src/Types.ts
@@ -514,7 +514,7 @@ class HeaderForm extends Component {
   // Update current track if the new tracks are valid
   // Otherwise check if the current bed file is a url, and if tracks can be fetched from said url
   // Tracks remain unchanged if neither condition is met
-  handleRegionChange = async (value, tracks, chunk) => {
+  handleRegionChange = async (value) => {
     // After user selects a region name or coordinates,
     // update path, region, and associated tracks(if applicable)
 
@@ -528,6 +528,24 @@ class HeaderForm extends Component {
       coords = this.getRegionCoords(value);
     }
     this.setState({ region: coords });
+    
+    let coordsToMetaData = {}
+
+    // Construct a concatenated string of possible coords
+    // Set relative meta data to each coord
+    if (this.state.regionInfo && !isEmpty(this.state.regionInfo)) {
+      for (const [index, path] of this.state.regionInfo["chr"].entries()) {
+        const pathWithRegion = path + ":" + this.state.regionInfo.start[index] + "-" + this.state.regionInfo.end[index];
+        coordsToMetaData[pathWithRegion] = {
+          "tracks": this.state.regionInfo.tracks[index],
+          "chunk": this.state.regionInfo.chunk[index]
+        }
+      }
+    }
+
+    // Set to null if any properties are undefined
+    const tracks = coordsToMetaData?.[coords]?.tracks ?? null;
+    const chunk = coordsToMetaData?.[coords]?.chunk ?? null;
 
     // Override current tracks with new tracks from chunk dir
     if (tracks) {
--- a/src/components/HeaderForm.js
+++ b/src/components/HeaderForm.js
@@ -10,7 +10,7 @@ import ExampleSelectButtons from "./ExampleSelectButtons";
 import RegionInput from "./RegionInput";
 import TrackPicker from "./TrackPicker";
 import BedFileDropdown from "./BedFileDropdown";
-import { parseRegion, stringifyRegion } from "../common.mjs";
+import { parseRegion, stringifyRegion, isValidURL } from "../common.mjs";
 
 
 // See src/Types.ts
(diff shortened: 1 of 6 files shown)


src/components/HeaderForm.js
```diff
@@ -527,6 +533,20 @@ class HeaderForm extends Component {
...
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ bedURL: this.state.bedFile, chunk: chunk }),
+      });
+
+      // Replace tracks if request returns non-falsey value
+      if (json.tracks) {
+        this.setState({ tracks: json.tracks });
```
adamnovak, line 548: It's possible that the `getChunkTracks` request comes back after the user has already swapped to a different BED region. We only want to apply the tracks from the server if the region we asked about is still the selected one.

I think we want to do something like a setState function here, so we can see the state that's current when the result arrives from the server:
```suggestion
        this.setState((laterState) => {
          if (laterState.region === coords) {
            // The user still has the same region selected, so apply the tracks we now have
            return { tracks: json.tracks };
          }
          // Otherwise, don't apply the downloaded tracks, because they are no longer relevant.
          // TODO: Save the downloaded tracks in case the user selects the region again?
        });
```

src/server.mjs
```diff
@@ -1360,6 +1367,47 @@ const timeoutController = (seconds) => {
...
+    // Attempt to read tracks.json and covnert it into a tracks object
+    if (fs.existsSync(track_json)) {
+      // Create string of tracks data
+      const string_data = fs.readFileSync(track_json);
+
+      // Convert to object container like the client component prop types expect
+      tracks = JSON.parse(string_data);
+    }
+
+    return tracks;
```
adamnovak, line 1390: This all looks 1 level too indented.

src/server.mjs
```diff
@@ -1360,6 +1367,47 @@ const timeoutController = (seconds) => {
...
+api.post("/getChunkTracks", (req, res, next) => {
+  console.log("received request for chunk tracks");
+  if (!req.body.bedURL || !req.body.chunk) {
+    throw new BadReqeustError("Invalid request format", req.body.bedURL, req.body.chunk);
+  }
+  let promise = (async () => {
+    // tracks are falsy if fetch is unsuccessful
+    const tracks = await getChunkTracks(req.body.bedURL, req.body.chunk);
+    res.json({ tracks: tracks });
+  })();
```
adamnovak, line 1404: This asynchronous operation also needs to hold a reader lock on the upload/download directories. Otherwise it isn't safe to call any of the download-and-read functions.

src/components/HeaderForm.js
```diff
@@ -527,6 +533,20 @@ class HeaderForm extends Component {
     if (tracks) {
       this.setState({ tracks: tracks });
       console.log("New tracks have been population");
+    } else if (isValidURL(this.state.bedFile)){
+      // Try to retrieve tracks if the current bed file is an URL
```
adamnovak, line 537: It looks like we're assuming that, if the BED file is not a URL and is actually mounted on the server, it also doesn't reference any chunks by URL and the server will always have sent us full track data for every region.

I don't think we should make that assumption on the client; we should be able to ask for the tracks for a BED region whenever we don't have them.

src/server.mjs
```diff
@@ -265,6 +265,31 @@ async function getChunkedData(req, res, next) {
...
+  } catch (e) {
+    // Whatever went wrong in the parsing, it makes the request bad.
+    throw new BadRequestError("Wrong query: " + e.message + " See ? button above.");
+  }
+
+  // There's a chance this request was sent before the proper tracks were fetched
+  // This can happen when the bed file is a url and track names need to be downloaded
+  // TODO: There's also a chance the user decided to use different tracks than what was provided in the bedFile,
+  // maybe we shouldn't always replace tracks if found?
+  if (req.body.bedFile && isValidURL(req.body.bedFile)) {
```
adamnovak, line 281: It looks like we're assuming that the client always will send the right tracks if their BED file is not a URL. I think we should run this check whether the BED is a URL or not.

src/server.mjs
```diff
@@ -265,6 +265,31 @@ async function getChunkedData(req, res, next) {
...
+    const chunk = await getChunkName(req.body.bedFile, parsedRegion);
+    const fetchedTracks = await getChunkTracks(req.body.bedFile, chunk);
+
+    // We're always replacing the given tracks if we were able to find tracks from the bed file
+    if (fetchedTracks) {
+      console.log("Using new fetched tracks", JSON.stringify(req.body.tracks));
+      req.body.tracks = fetchedTracks;
```
adamnovak, line 288: I think we need to do something more complex here to avoid clobbering the color information, which is stored inside the tracks. If the client hasn't sent tracks, we should fill in the tracks with their default color settings. If the client has sent tracks *and* has set up their own color selections, we need to retain the client's color choices.

I think if we don't, the client will apply whatever tracks we set here, when the request comes back, and then display with the default colors instead of the user-selected colors, right?

src/components/RegionInput.js
adamnovak, line 1: I don't think the `RegionInput` should be responsible for managing all the region metadata and passing it up to `HeaderForm`, like we have it now. So I think instead of adding `optionToChunk` to this component, we want to remove `optionToTrack`.

Then `handleRegionChange` can get back the responsibility for figuring out what track data to use with the current region, which means it will be able to keep a cache of track data for regions, and use track sets from there when they are present, or else request the tracks and put them in the cache when they come back.

If we leave `RegionInput` in charge of telling the `HeaderForm` when a region has a known track set, but then `HeaderForm` is responsible for second-guessing that determination and asking the server, the responsibility is split weirdly between the two components.

src/server.mjs
```diff
@@ -1446,16 +1494,12 @@ async function getBedRegions(bed) {
...
 
+      // See if we have downloaded tracks.json in a previous instance
       let track_json = path.resolve(chunk_path, "tracks.json");
 
-      let tracks_array = [];
-
-      // If json file specifying the tracks exists
+      // If json file specifying the tracks exists, pass its information into a tracks object
+      // future selection of this region won't re-fetch tracks.json
       if (fs.existsSync(track_json)) {
```
adamnovak, line 1503: We might want to think about not sending any track data with the `getBedRegions` response (in order to simplify things and reduce the number of distinct ways you have to worry about sending track data).

Or maybe only sending track data when there are just a few regions?

If somebody makes a *really* big BED, and mounts it onto the server, we don't want to have to be jumping all over the disk opening thousands of JSON files to answer the request, especially if the user is only actually going to need a few of them, one at a time.

adamnovak (changes requested):
This looks like it's mostly going to work!

But I think we have it set up for very different things happening, in terms of when track data is delivered to the client and when the client is assumed to have it, depending on whether the BED is coming from a URL or not. I think we should try and make the same flows happen in both cases, as much as we can.

Also, I think the `RegionInput` is developing an extra responsibility as a database of region metadata and track information. I think we should rip that back out and put it somewhere else, either into `HeaderForm` (where the database of track metadata is actually stored as state, and where track metadata for regions is actually fetched from the server), or maybe somehow into a free-floading, non-React-component class (if having it in `HeaderForm` makes `HeaderForm` too complicated).

ducku:
Failing

> config-global.mjs loaded before either config-client.js or config-server.mjs 

I'm not too sure why this happens.

adamnovak:
@ducku This is a new check I added. Since `common.mjs` needs the config, but since the code to load the config on the client can't be imported on the server and visa versa, and since imports can't be conditional, you need to import `config-client.js` before `common.mjs` in client code, and you need to import `config-server.mjs` before `common.mjs` in server code.

adamnovak:
I couldn't figure out a better way to write code that needs access to the config and also needs to be portable between the client and server.

src/components/RegionInput.js
```diff
@@ -3,10 +3,10 @@ import PropTypes from "prop-types";
 import TextField from "@mui/material/TextField";
 import Autocomplete from "@mui/material/Autocomplete";
 import FormHelperText from "@mui/material/FormHelperText";
+import { isEmpty } from "../common.mjs";
```
adamnovak, line 6:
```suggestion
import "../config-client.js";
import { isEmpty } from "../common.mjs";
```

adamnovak, line 6:
```suggestion
```

The config isn't actually needed here anymore.

src/components/RegionInput.js
```diff
@@ -6,7 +6,6 @@ import FormHelperText from "@mui/material/FormHelperText";
 
 // RegionInput: The path and region input box component
 // Responsible for selecting the path/chr and segment of data to look at
-const isEmpty = (obj) => Object.keys(obj).length === 0;
```
adamnovak, line 9: I think you need to keep this in for the component to work.

src/server.mjs
```diff
@@ -283,8 +283,30 @@ async function getChunkedData(req, res, next) {
 
     // We're always replacing the given tracks if we were able to find tracks from the bed file
     if (fetchedTracks) {
-      console.log("Using new fetched tracks", JSON.stringify(req.body.tracks));
-      req.body.tracks = fetchedTracks;
+      // Color Settings are retained from the initial request
+      // if newly fetched tracks have matching file names
+      // Store current colors and file names
+      const fileToColor = new Map();
```
adamnovak, line 289: We're sure that we're not going to have multiple tracks with the same filename?

Can we not use the track key because we want to think about tracks being added/removed/reordered?

src/components/RegionInput.js
adamnovak: Oh I guess we want to import the new `isEmpty` function from `common.mjs`. Which we can do if we import `config-client.js` first.

adamnovak (approved):
This looks pretty good, once the right stuff is imported to make it run.

I'm a little worried about forbidding duplicate filenames in pre-made chunks' track sets, but probably that will be fine?

What kind of snapping back to the correct tracks are we going to expect to see if the user changes the tracks and rerenders a premade BED region?

PR #369, opened 2023-11-26
Display Region Description as Label and Tooltip

### LINKED ISSUE #361: Get the region description to be visible on the page when looking at the region
Right now, when you select a region by description from the dropdown, you can't actually see its description anymore, even if you git Go to bring it up.

We should display the description of a BED region on the page somewhere, when the user is looking at that region. It could appear in the region box along with the coordinates, or it could appear as text on the page (which would be easier to implement).

ducku:
Closes #361


ducku:
Changed tooltip position to no longer block dropdown, reverted label to "Region".

Region_Desc

adamnovak (approved):

PR #374, opened 2023-11-30
Eliding small snarls

### LINKED ISSUE #196: Eliding small snarls
@benedictpaten has the idea that maybe we could have a way to remove small snarls (e.g. single base indels and SNPs) from the tube map, to get a more legible higher-level zoomed-out view.

This might involve having and then modifying the snarl structure of the graph, and might require support on the `vg` side.

shreyasun:
Fixes #196


adamnovak:
@shreyasun It looks like you need to add `&simplify=false` to the correct copied links in the link copying tests. Can you do that?

public/help/help.md
```diff
@@ -27,5 +27,6 @@ The following procedure describes adding and updating settings of custom tracks.
    * A start position and a distance (e.g. "chr1:1+100")
    * A node ID anchor and a distance (e.g. "node:100+10")
 ![Region Input Options](helpGuideImages/img8.png)
-9. Click Go to see the selected tracks render in the visualization area.
+4. There is an option to click on the "Simplify Off" button to enable vg simplify, which would remove eliding small snarls in the graph. This would only work for BED files or graph tracks (not reads or haplotypes).
```
adamnovak, line 30: This makes it sound like you can turn on simplification for all BED file chunks. It also reads like the user should expect to see a non-functional simplify button sometimes, when actually the button only appears when it is going to work.

I think we want to communicate that the button is not always visible (so people aren't surprised when they don't see it), and that people should expect to see it only when there aren't any reads to be displayed.

I think it *does* work with haplotype tracks; the haplotype information goes into `vg chunk` and the `vg simplify` pass happens after that.

src/common.mjs
```diff
@@ -111,6 +111,16 @@ export function defaultTrackColors(trackType){
   }
 }
 
+/* Function to determine if reads are added, where tracks is an object */
```
adamnovak, line 114: Tracks isn't just any object, right? It is meant to be an object full of track objects (which is a type we've defined).

And by "reads are added", we mean that we're checking to see if any of the tracks is a read track. I'm not sure "added" is the right word here; we're checking to see if they are present.

src/components/CopyLink.js
```diff
@@ -81,9 +81,16 @@ export const urlParamsToViewTarget = (url) => {
...
-  // some point.
+  // Ensures that the simplify field is a boolean, as the qs module can't tell 
+  // the difference between false and "false" 
+  if (result != null){
+    let resultSimplify = result.simplify;
+    if (resultSimplify === "true"){
+      result.simplify = true;
+    } else if (resultSimplify === "false"){
+      result.simplify = false;
+    }
```
adamnovak, line 92: I think the `resultSimplify` variable could be eliminated here to improve the code.

src/components/HeaderForm.js
```diff
@@ -713,6 +718,11 @@ class HeaderForm extends Component {
     };
   };
 
+  /* Function for simplify */
+  enableSimplify = () => {
```
adamnovak, line 722: This doesn't actually "enable" simplification; it toggles it. I think toggling it is the correct behavior so the function name should be changed.

src/components/HeaderForm.js
```diff
@@ -713,6 +718,11 @@ class HeaderForm extends Component {
     };
   };
 
+  /* Function for simplify */
```
adamnovak, line 721: This comment doesn't do very much to explain the function. It should say what the function needs to do. Or you could say this is the setter function for the simplify two-way binding to the button and let that imply what it is supposed to do.

src/components/HeaderForm.js
```diff
@@ -752,6 +762,9 @@ class HeaderForm extends Component {
     const examplesFlag = this.state.dataType === dataTypes.EXAMPLES;
     const viewTargetHasChange = !viewTargetsEqual(this.getNextViewTarget(), this.props.getCurrentViewTarget());
 
+    const ifReadsExist = readsExist(this.state.tracks);
```
adamnovak, line 765: It looks like this variable is sued exactly once, so maybe the function call should just be made where the variable is used?

Or if you need the variable to simplify the complicated logic where it is used, maybe we could move the variable closer to the logic it goes with?

src/server.mjs
```diff
@@ -664,13 +730,64 @@ async function getChunkedData(req, res, next) {
...
+
+    // vg simplify
+    if (req.simplify){
+      vgSimplifyCall.on("error", function (err) {
+        console.log(
+          "Error executing " +
+            VG_PATH +
+            "vg " +
+            "simplify " + 
+            "- " +
```
adamnovak, line 757: This doesn't accurately report the command that was called; when we run `vg simplify` here we give it a filename as an argument and not `-`.

src/server.mjs
```diff
@@ -664,13 +730,64 @@ async function getChunkedData(req, res, next) {
...
+
+      vgSimplifyCall.stdout.on("data", function (data) {
+        vgViewCall.stdin.write(data);
+      });
+
+      vgSimplifyCall.on("close", (code) => {
+        console.log(`vg simplify exited with code ${code}`);
+        vgViewCall.stdin.end();
+        if (code !== 0) {
+          console.log("Error from " + VG_PATH + "vg " + "simplify - ");
```
adamnovak, line 781: Here also we're incorrectly claiming that we called `vg simplify -`.

adamnovak (changes requested):
This looks like it should work great, but some of the comments, error messages, and documentation need to be improved before we merge it.

Also, there are some places where the code should be simplified.

adamnovak (approved):
I think my comments were addressed.

PR #376, opened 2023-12-02
Improve Region Dropdown Selections

### LINKED ISSUE #362: Improve region dropdown display for BED regions
- [ ] Right now, BED regions with no descriptions appear twice in the dropdown, once as coordinates and once as "Region {coordinates}". This is confusing. We should not display dropdown options for these auto-generated descriptions.

- [ ] For regions with manually set descriptions, the coordinates and the description also appear as multiple entries in the dropdown, although they are semantically one region. We should display them as a single item if possible, maybe by using some sort of structured item support in the control, or maybe by concatenating them together and using the ability of the dropdown to display a label while holding a distinct value.

- [ ] There are `none:` entries in the region dropdown; they should be eliminated.

ducku:
Use label system to display both path with region and description while ignoring description upon onChange
Region_Dropdown

Closes #362

### DIFF
--- a/src/components/RegionInput.js
+++ b/src/components/RegionInput.js
@@ -40,7 +40,7 @@ export const RegionInput = ({
   }
 
   // Autocomplete selectable options
-  const displayRegions = [...pathsWithRegion, ...pathNamesColon].filter(option => option.label !== "none:");
+  const displayRegions = [...pathsWithRegion, ...pathNamesColon];
 
   let descLabel = "Region";
   if (regionToDesc.get(region)) {
--- a/src/components/RegionInput.js
+++ b/src/components/RegionInput.js
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import TextField from "@mui/material/TextField";
 import Autocomplete from "@mui/material/Autocomplete";
 import FormHelperText from "@mui/material/FormHelperText";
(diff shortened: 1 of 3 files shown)


src/components/RegionInput.js
```diff
@@ -19,52 +20,68 @@ export const RegionInput = ({
...
     }
-
-    // Add descriptions
-    pathsWithRegion.push(...regionInfo["desc"]);
-
   }
 
   // Autocomplete selectable options
-  const displayRegions = [...pathsWithRegion, ...pathNamesColon];
+  const displayRegions = [...pathsWithRegion, ...pathNamesColon].filter(option => option.label !== "none:");
```
adamnovak, line 43: I don't think this is the Right Way to get rid of this extra option, because the code does not convince me that we actually know why it is showing up.

Either we should get rid of it where it is actually being generated, or we should have a comment here explaining why it is there and why this actually is the best time to remove it.

adamnovak (changes requested):
This looks good and seems to work properly when I run it, except I don't like the implementation of getting rid of the `none:` option. We're filtering it out instead of working out why it is getting generated and making that not happen.

Also the PR probably needs to be updated with the merge of the other region dropdown PR.

adamnovak (approved):

Issue #354, opened 2023-10-11
Make the left and right buttons go to the previous/next BED region when a BED file is selected


adamnovak:
If a BED subregion contains a `tracks.json` that doesn't point at files the server is allowed to use (because e.g. it was extracted from a larger graph that is not available on the web), we should figure this out and stop the user from trying to use that file.

For example, we should turn off the left and right buttons, because we know we won't be able to fetch a different set of coordinates (or we could make them go left/right to the next/previous region *in the BED file*).

We should also stop the user from going to coordinates not represented in the BED with a premade chunk, if they don't actually have a set of tracks that we could fetch the region from.

(Maybe we want to handle this by allowing empty/null track files in `tracks.json` in a BED region? Then we can use null by default in the extraction scripts, and you could make sure to fill in the paths manually if you want a BED with pre-extracted regions but also a real backing graph.)

adamnovak:
I need to add a sequence of steps that can be used to reproduce this issue before it can actually be fixed.

adamnovak:
To reproduce this issue:

1. Pull in changes from the `master` branch now that https://github.com/vgteam/sequenceTubeMap/pull/375 is merged.
2. Start the tube map.
3. Set `Data` to `custom`.
4. Set `BED file` to `no_source.bed`.
5. Set `Region` to `ref:2000-3000`.
6. Click `Go`.
7. Click the left button or the right button.

You will get an error:
```
vg chunk failed: error:[vg chunk] unable to load graph / xg index file exampleData/nonexistent.vg 
```

Instead of letting this happen, we should probably look in the BED file for a next or previous region, and jump to that instead. In this particular BED file there is just one region, so we should grey out the next/previous buttons since we are at the first region and we are at the last region.

We also get the same error if, having visited `ref:2000-3000`, you type some other region manually like `ref:123-456` and hit `Go`. Instead of letting the server complain that the graph file doesn't exist, we might want to somehow know that there's no full backing graph and disable free region entry, and make you pick from the available regions. To make that work, we might have to change the `tracks.json` format a little bit, and maybe use an empty track filename or something when the full graph isn't meant to be available, and then detect that and change the region select behavior.

adamnovak:
For implementation, so far we have a function from region string to index in the collection of available regions, if present.

The buttons need to:
1. Go from current region string to region index number (with function that is written)
[...]

PR #380, opened 2023-12-10
Implement highlightable description

### LINKED ISSUE #378: Move region description tooltips to highlightable text on the page
Right now we display region descriptions as tooltips. But the Lancet folks want to put e.g. gene names in there which the user might want to copy-paste. So we should put the region description, when set, as text on the page, maybe between the buttons and the main view. If there isn't a description, we shouldn't show an empty space.

ducku:
Puts Description, if available, as text below Go button.

Closes #378

### DIFF
--- a/src/components/RegionInput.js
+++ b/src/components/RegionInput.js
@@ -20,7 +20,9 @@ export const RegionInput = ({
   // Generate autocomplete options for regions from regionInfo
   // Add : to pathNames
   console.log("rendering with pathnames: ", pathNames);
-  const pathNamesColon = pathNames.map((name) => name + ":");
+  const pathNamesColon = pathNames.map((name) => {
+    return {label: name + ":", value: name + ":"};
+  });
   const pathsWithRegion = [];
 
   const regionToDesc = new Map();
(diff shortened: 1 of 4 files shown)


src/components/RegionInput.js
```diff
@@ -52,14 +53,20 @@ export const RegionInput = ({
...
-          onInputChange={(event, newInputValue) => {
-            handleRegionChange(newInputValue);
+          onInputChange={(event, newInput) => {
+            let regionValue = newInput;
+            const regionObject = displayRegions.find((option) => option.label === newInput);
+            // If input is selected from one of the options
+            if (regionObject) {
+              regionValue = regionObject.value
+            }
+            handleRegionChange(regionValue, regionToDesc.get(regionValue));
```
adamnovak, line 69: I feel like the regionToDesc lookup might really want to be in `handleRegionChange`? So it doesn't need a new argument? But we kind of maintain the whole database of regions in RegionInput and it would be odd to do it twice.

I think Shreya is working on the left/right button jump-to-next-region thing in #354 though and that requires having more of an understanding of the region database in the header form, so I think this sort of thing is moving out of RegionInput in the medium term.

adamnovak (approved):
It looks like this should work. I think the extra argument to handleRegionChange is the wrong way to do it, but I think to do it the right way it has to be coordinated with @shreyasun and what she ends up doing for the left/right button work.

adamnovak:
Here's a screenshot of the text on the page, under the go button, with its heading: Screenshot 2023-12-12 at 12 27 37

It's tied to what region you have selected/typed, not to what region is currently displayed, which might be a little confusing? But it's probably fine.

PR #382, opened 2024-01-08
Added read path information to read info dialog

### LINKED ISSUE #363: Add read path information to read info dialog
In addition to the CIGAR string, the dialog that opens when you click on a read should include the path that the read takes through the graph's nodes. This will help with debugging unusual read layout rendering.

We might want to mimic the way GAF or GFA paths work, with `>1>2<3>4` style path strings.

ducku:
Closes #363

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -3918,6 +3918,26 @@ function trackDoubleClick() {
   createTubeMap();
 }
 
+// Takes a track and returns a string describing the nodes it passes through
+// In the format of >1>2<3>4, with the intergers being nodeIDs
+function getPathInfo(track) {
+  let result = "";
+  if (!track.sequence) {
+    return result;
+  }
+
+  for (const nodeID of track.sequence) {
+    // Node is approached backwards if "-" is present
+    if (Array.from(nodeID)[0] === "-") {
+      result = result.concat("<", nodeID.substring(1));
+    } else {
+      result = result.concat(">", nodeID);
+    }
+  }
+  
+  return result;
+}
+
 function trackSingleClick() {
   /* jshint validthis: true */
   // Get the track ID as a number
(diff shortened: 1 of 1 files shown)


src/util/tubemap.js
```diff
@@ -3918,6 +3918,26 @@ function trackDoubleClick() {
...
+// In the format of >1>2<3>4, with the intergers being nodeIDs
+function getPathInfo(track) {
+  let result = "";
+  if (!track.sequence) {
+    return result;
+  }
+
+  for (const nodeID of track.sequence) {
+    // Node is approached backwards if "-" is present
+    if (Array.from(nodeID)[0] === "-") {
```
adamnovak, line 3931: This seems like a weird way to see if the string starts with "-", even though it looks like [StackOverflow recommends it](https://stackoverflow.com/a/3427148), since this looks to convert the whole string to an array just to check the first character. I would probably try the `startsWith` method instead.
```suggestion
    if (nodeID.startsWith("-")) {
```

src/util/tubemap.js
```diff
@@ -3918,6 +3918,26 @@ function trackDoubleClick() {
...
+      result = result.concat("<", nodeID.substring(1));
+    } else {
+      result = result.concat(">", nodeID);
```
adamnovak, line 3934: This looks like it is going to be an O(n^2) algorithm when it doesn't need to be. Building the new version of `result` is going to copy the original contents of `result` into a new string that also has the `<`/`>` character and the node ID. But that copy takes some unit of time for each character in `result`, and as you make `result` longer the number of characters to be copied increases at each step.

You probably want to build an array of all the pieces you want in the final string, by mutably appending to the array with `push()` (which doesn't take longer to do as the array gets longer). Then you can use `.join("")` on the array to turn it into a combined string all at once, in linear time.

adamnovak (changes requested):
This looks like it should work, but I think the logic to build the string has a bigger `O()` runtime than is really justifiable.

adamnovak (approved):
This looks good to me, and I manually tested it.

If redundant nodes are off, we get the path through the modified graph and not the original graph before the node compression. I'm not entirely sure which way is better, so I guess we can ship this and change it if we decide the other way is actually better.

Issue #377, opened 2023-12-04
Refactor requests to the API so that they could be computed locally instead


adamnovak:
To implement #367, we are going to need to set things up so that sometimes, when you click "Go", instead of making an API request to `/getChunkedData`, we instead call some local JS/WASM code which makes its own web requests and/or operates on local files exposed to the browser, and eventually produces the data needed to render the tube map.

To support this, we should refactor the client-side code so that the current code that calls `/getChunkedData` is just one possible implementation of an interface.

So, for this issue, we should define a JavaScript `class` to represent that interface, and another `class` inheriting from the first that implements it by making requests to the tube map API server. The interface should cover:

1. Requesting the chunk data given the tracks, region, and other settings.
2. Taking a [File API](https://developer.mozilla.org/en-US/docs/Web/API/File_API) file and turning it into a URL to put in a track or use as a BED (i.e. doing the upload in the current implementation).
3. Getting the list of paths in a graph, given a graph URL (which might be a URL from step 2).
4. Getting track information for a BED file by URL (which also might be a URL from step 2).

Then we should replace the existing API code with code to instantiate an instance of this class, and to make API calls through the current implementation of the interface.

adamnovak:
This could look like taking existing calls to `fetchAndParse` with particular paths based on `apiUrl`, and changing them to be calls to a method of this new interface, and then passing an instance of the interface down in the React props instead of `apiUrl`.

So a [call like](https://github.com/vgteam/sequenceTubeMap/blob/bc8040fc50be085557d4cde0fecbba531db8248a/src/components/HeaderForm.js#L297C7-L303C10):

```
const json = await fetchAndParse(`${this.props.apiUrl}/getFilenames`, {
        signal: this.cancelSignal, // (so we can cancel the fetch request if we will unmount component)
        method: "GET",
        headers: {
          "Content-Type": "application/json",
        },
      });
```

Would become something like:
```
const json = await this.props.api.getFilenames(this.cancelSignal);
```

Or [with an argument](https://github.com/vgteam/sequenceTubeMap/blob/bc8040fc50be085557d4cde0fecbba531db8248a/src/components/HeaderForm.js#L354-L361):

```
const json = await this.props.api.getBedRegions({ bedFile }, this.cancelSignal);
```

We can send the arguments either always as an object, or for one-argument methods we could decide that we don't actually want the wrapping object and generate it inside the implementation if needed.

adamnovak:
[...]

PR #384, opened 2024-01-08
Add sketch of WASM integration

adamnovak:
This pulls in a dependency on an NPM module with the WASM blobs for doing API requests locally, and also adds a skeleton of an API implementation and the beginning of tests for it.

I still can't actually load the WASM blob; I think I am going to need to swap out some code between running on Jest and building with Webpack somehow, to get the bytes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
+        "@bjorn3/browser_wasi_shim": "^0.2.17",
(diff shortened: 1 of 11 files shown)


adamnovak:
In addition to the WASM stuff that doesn't actually work yet, this has a fix for the upload and websocket parts of the API refactor in #377.

adamnovak:
I got the WASM stuff to load and execute *something* in the browser. It still doesn't actually implement the API but I think we can fill that in piece by piece.

PR #386, opened 2024-01-14
Issues/364 prevent reads shuffle

### LINKED ISSUE #364: Reads skipping multiple nodes do a nonsensical shuffle
With multiple GAM files (?), when reads skip over multiple nodes, their vertical order when skipping the first node gets lost when skipping subsequent nodes, and they do this absurd complete shuffle.

Image

The reads should flow through coherently.

I observed this in `chr6:82003584-82004084` from Rajeeva's data set he sent over for debugging looping reads.

ducku:
Reads outside of nodes are sorted by order, then a previous y value. However, consecutive reads outside of nodes have their y values computed simultaneously, resulting in the previous y value still being null. This resulted in consecutive read segments outside of nodes that have the same order value being randomly sorted.

Fix: Go back through the path and find the last computed y value(from the last visited node), and use that to sort segments.

Closes #364 

`chr6:82003584-82004084`

Capture

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -678,7 +678,7 @@ function placeReads() {
         // use previous y value from last segment with node instead 
         let previousValidY = null;
         let lastIndex = pathIdx - 1;
-        while (!previousValidY && lastIndex >= 0) {
+        while (previousValidY === null && lastIndex >= 0) {
           previousValidY = reads[idx].path[lastIndex].y;
           lastIndex = lastIndex - 1;
         }
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -688,8 +688,8 @@ function placeReads() {
       }
     });
   });
-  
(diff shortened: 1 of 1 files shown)


src/util/tubemap.js
```diff
@@ -674,14 +674,23 @@ function placeReads() {
   reads.forEach((read, idx) => {
     read.path.forEach((element, pathIdx) => {
       if (!element.hasOwnProperty("y")) {
+        // previous y value from pathIdx - 1 might not exist yet if that segment is also without node 
+        // use previous y value from last segment with node instead 
+        let previousValidY = null;
+        let lastIndex = pathIdx - 1;
+        while (!previousValidY && lastIndex >= 0) {
```
adamnovak, line 681: I think testing for truthiness here excludes y=0 without a real justification.

```suggestion
        while (previousValidY === null && lastIndex >= 0) {
```

adamnovak (approved):
This looks good, I am just going to slightly tinker with the logic.

PR #387, opened 2024-01-23
Fix left right buttons

### LINKED ISSUE #354: Make the left and right buttons go to the previous/next BED region when a BED file is selected

shreyasun:
Fixes #354


adamnovak (approved):
This looks good to me.

PR #389, opened 2024-01-28
Improve Read Cycle Layout

### LINKED ISSUE #321: Layout of reads going in cycles is bad
With the high coverage Lancet graphs, there can be a lot of reads going in cycles through the first and last nodes of a graph.

If they are actually meant to be doing this (which I'm not sure they are), the layout is bad. They don't stay in a consistent order among their multiple visits to the node, or in the order they decide to hang a right and go downwards. So they shuffle at each 90 degree turn involved in going around and it is hard to see which one is which.

We should get them to move in nice GBWT-like bands all the way around these cycles.

Also, the space needed for them to extend out and turn downwards is not accounted for; other nodes can lay out there, meaning the reads are making turns under other nodes. We should push the other columns of nodes rightwards enough to make space for all the reads to turn.

Here's a screenshot of these problems:

Image

And [a link to the view to reproduce them](http://127.0.0.1:3001?tracks[1][trackType]=graph&tracks[1][trackColorSettings][mainPalette]=blues&tracks[1][trackColorSettings][auxPalette]=reds&tracks[1][trackColorSettings][colorReadsByMappingQuality]=false&tracks[1][trackFile]=chr6_29925378_29927878__HCC1395.giraffe.gbz&tracks[2][trackType]=haplotype&tracks[2][trackColorSettings][mainPalette]=plainColors&tracks[2][trackColorSettings][auxPalette]=reds&tracks[2][trackColorSettings][colorReadsByMappingQuality]=false&tracks[2][trackFile]=chr6_29925378_29927878__HCC1395.giraffe.gbz&tracks[3][trackType]=read&tracks[3][trackColorSettings][mainPalette]=reds&tracks[3][trackColorSettings][auxPalette]=reds&tracks[3][trackColorSettings][colorReadsByMappingQuality]=false&tracks[3][trackFile]=Normal_HCC1395.chr6_29925378_29927878.sorted.gam&tracks[4][trackType]=read&tracks[4][trackColorSettings][mainPalette]=blues&tracks[4][trackColorSettings][auxPalette]=blues&tracks[4][trackColorSettings][colorReadsByMappingQuality]=false&tracks[4][trackFile]=Tumor_HCC1395.chr6_29925378_29927878.sorted.gam&bedFile=none&region=ref0%3A1-100&dataPath=mounted&dataType=mounted%20files) if you have the Lancet data from the "complex" case.

ducku:
Segments cycling on the first node had strange interactions since segments classified as "outgoing"(if its pathIndex == 0) had different methods of sorting.

Changes are made to how reads are placed outside of nodes when a cycle is detected, and inside of nodes when reads have the same source and destination Node.

- A new tiebreaker is placed in sortByGoingTo for reads that start and end in the same node. Now they're placed based on the endmost node they were placed in, working its way back if there are no y values. Since these reads start and end in the same node, this should start the reads(in the first node) in the same order as the other nodes.

- Segments outside of nodes now detects cycles, defined as the previous and next segments are in the node and going in the same direction. Segments outside of nodes and are between cycles now sort in reverse to keep the layout clean.

- Extra space in between nodes are calculated after reads are added into the tracks variable.

Cycle1

Cycle2

Closes #321

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -685,25 +685,31 @@ function placeReads() {
           lastIndex = lastIndex - 1;
         }
 
-        // sometimes, elements without nodes are between 2 segments going to the same node, from the same direction
-        // this means we're looping back to the same node, which constitutes a different sorting priority
+        // sometimes, elements without nodes are between 2 segments going to a node we've already visited, from the same direction
+        // this means we're looping back to a node we've already been to, and we should sort in reverse
         
-        let previousNodeOrder = reads[idx].path[pathIdx - 1]?.order;
         let nextNodeOrder = reads[idx].path[pathIdx + 1]?.order;
+        let previousNodeOrders = reads[idx].path.map(p => p.order).slice(0, pathIdx);
 
-        let previousNodeIsForward = reads[idx].path[pathIdx - 1]?.isForward;
-        let nextNodeIsForward = reads[idx].path[pathIdx + 1]?.isForward
 
         // if the previous segment and the next segment is going to the same node
-        let sameNode = previousNodeOrder !== null && nextNodeOrder !== null && previousNodeOrder === nextNodeOrder;
-        // if the previous semgne tnad the next segment is going in the same direction
-        let sameDirection = previousNodeIsForward !== null && nextNodeIsForward !== null && previousNodeIsForward === nextNodeIsForward
+        let nextNodeVisited = previousNodeOrders !== null && nextNodeOrder !== null && previousNodeOrders.includes(nextNodeOrder);
 
         let betweenCycle = false;
-        // if the previous segment and the next segment is going to the same node
-        if (sameNode && sameDirection) {
-          betweenCycle = true;
+        if (nextNodeVisited) {
+          // check the forward status of the last time we visited the node we're about to visit
+          let previousNodeIsForward = reads[idx].path[previousNodeOrders.lastIndexOf(nextNodeOrder)]?.isForward;
+          let nextNodeIsForward = reads[idx].path[pathIdx + 1]?.isForward
+
+          // if the next segment and the one we've previously visited is going in the same direction
+          let sameDirection = previousNodeIsForward !== null && nextNodeIsForward !== null && previousNodeIsForward === nextNodeIsForward
+
+          // we've already visited the next node and we're going the same direction as the last time we've visited it
+          if (sameDirection) {
+            betweenCycle = true;
+          }
         }
+
 
         elementsWithoutNode.push({
           readIndex: idx,
@@ -887,10 +893,28 @@ function compareReadOutgoingSegmentsByGoingTo(a, b) {
   let readIndexA = a[0];
   let readIndexB = b[0];
 
-  // Segments are first sorted by the the node they end on,
-  // then by a coming-from tiebreaker,
+
+  // Segments are first sorted by the y value of their last node,
+  // then by the node they end on,
   // then by length in final node
-  // TODO: Perhaps we should make an exception to first sort by a coming-from when a cycle is involved
+  let previousValidYA = null;
+  let previousValidYB = null;
+  let lastPathIndexA = reads[readIndexA].path.length - 1;
+  let lastPathIndexB = reads[readIndexB].path.length - 1;
+  while ((previousValidYA === null || !previousValidYA) && lastPathIndexA >= 0) {
+    previousValidYA = reads[readIndexA].path[lastPathIndexA].y;
+    lastPathIndexA -= 1;
+  }
+  while ((previousValidYB === null || !previousValidYB) && lastPathIndexB >= 0) {
+    previousValidYB = reads[readIndexB].path[lastPathIndexB].y;
+    lastPathIndexB -= 1;
+  }
+  
+  if (previousValidYA && previousValidYB) {
+    return previousValidYA - previousValidYB;
+  }
+
+  // Couldn't find a valid y value for at least one of the reads, sort by which node reads end on
   let nodeA = nodes[reads[readIndexA].path[pathIndexA].node];
   let nodeB = nodes[reads[readIndexB].path[pathIndexB].node];
   // Follow the reads' paths until we find the node they diverge at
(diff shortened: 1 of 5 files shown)


src/util/tubemap.js
```diff
@@ -682,10 +684,32 @@ function placeReads() {
           previousValidY = reads[idx].path[lastIndex].y;
           lastIndex = lastIndex - 1;
         }
+
+        // sometimes, elements without nodes are between 2 segments going to the same node, from the same direction
+        // this means we're looping back to the same node, which constitutes a different sorting priority
```
adamnovak, line 689: I'm not sure "constitutes" is the right word here. Do you mean we want to use a different sorting in this case?

src/util/tubemap.js
```diff
@@ -488,7 +488,7 @@ function createTubeMap() {
   generateLaneAssignment();
 
   if (config.showExonsFlag === true && bed !== null) addTrackFeatures();
-  generateNodeXCoords();
+  //generateNodeXCoords();
```
adamnovak, line 491:
```suggestion
```
We don't need this here anymore since we moved this call.

src/util/tubemap.js
```diff
@@ -682,10 +684,32 @@ function placeReads() {
...
+        // if the previous segment and the next segment is going to the same node
+        let sameNode = previousNodeOrder !== null && nextNodeOrder !== null && previousNodeOrder === nextNodeOrder;
+        // if the previous semgne tnad the next segment is going in the same direction
+        let sameDirection = previousNodeIsForward !== null && nextNodeIsForward !== null && previousNodeIsForward === nextNodeIsForward
+
+        let betweenCycle = false;
+        // if the previous segment and the next segment is going to the same node
+        if (sameNode && sameDirection) {
+          betweenCycle = true;
+        }
```
adamnovak, line 706: I think we might want to use this reverse sort order logic in more cases than *just* the one where we visit the same node in the same orientation before and after the no-node section.

I think we want to have the same behavior even if that one node that we loop around is instead divided up into two or more nodes that we loop around as a group. Like it the read visits nodes 1, 2, 1, 2, 1, 2. I'm not sure of the best way to detect that we are visiting no node because we need to get back to an earlier X to complete a cycle, though.

adamnovak (changes requested):
This looks pretty good, and way better than before.

But I'm concerned that it's specialized for one-node cycles, when probably we need to support cycles of multiple nodes.

Also, what's going on in the first screenshot with the innermost three reads in the spiral below the node? They don't seem to be spiraling in the same order as the other ones, and so they end up crossing over each other for no apparent reason. It's not a big problem for these 3 reads, but I think a similar problem is happening with more reads in the second screenshot, and it makes the visualization messier there.

ducku:
Changing to sort the first segment by existing y values first works better on this set of data, and I think it would work better overall. 

I'm currently trying to make an example dataset showcasing this.

Cycle1
Cycle2

src/util/tubemap.js
```diff
@@ -676,12 +677,57 @@ function placeReads() {
...
+        } else if (nextNodeJustVisited) {
+          // the segment doesn't have to loop if going in opposite direction from the node
+          let previousNodeIsForward = previousPathWithNode?.isForward;
+          let nextNodeIsForward = nextPathWithNode?.isForward;
+
+          // if the next segment(to a node) and previousl segment(from a node) is going in the same direction
+          let sameDirection = previousNodeIsForward !== null && nextNodeIsForward !== null && previousNodeIsForward === nextNodeIsForward
+          
+          // we're visiting the same node we just visited from the same direction
+          if (sameDirection) {
```
adamnovak, line 724: Shouldn't we just put the conditions in the ifs, if we're only going to fill in the variable to immediately condition on it? It's useful to have the variable name for documentation, but we could also just have a comment.

src/util/tubemap.js
```diff
@@ -676,12 +677,57 @@ function placeReads() {
...
+            reads[idx].path[pathIdx].betweenCycle = true;
+          }
+        }
```
adamnovak, line 727: All this stuff is one big boolean function to fill in `betweenCycles`; I feel like it might be able to be simplified somehow.

If either side doesn't exist or has `order` (or `isForward`) not set, the answer is `false`. Otherwise, we are `true` if the orders and the orientations are equal, or if the previous order exceeds the next order.

You could turn that into one expression with comments. If we don't have to deal with real nulls in there, that would be something like:
```
// If there is a next and previous, we can be between cycles
let betweenCycles = (nextPathWithNode && previousPathWithNode &&
  // If we go back to the same node in the same orientation
  ((nextPathWithNode.order == previousPathWithNode.order && nextPathWithNode.isForward == previousPathWithNode.isForward) ||
  // Or we go back to a previous node in any orientation
  (nextPathWithNode.order < previousPathWithNode.order)));
```

src/util/tubemap.js
```diff
@@ -676,12 +677,57 @@ function placeReads() {
...
+        while ((nextPathWithNode?.node === null || !nextPathWithNode?.node) && nextPathIndex < reads[idx].path.length) {
+          nextPathWithNode = reads[idx].path[nextPathIndex];
+          nextPathIndex = nextPathIndex + 1;
+        }
+
+        let nextNodeOrder = nextPathWithNode?.order;
+        let previousNodeOrder = previousPathWithNode?.order;
+
+        // if the previous segment and the next segment is going to the same node
+        let nextNodeJustVisited = previousNodeOrder !== null && nextNodeOrder !== null && previousNodeOrder === nextNodeOrder;
```
adamnovak, line 704: Are there actual `null`s stored in `order` and/or `isForward` on the visits? Or should these be checking for `undefined` from the optional chaining `?.` instead?

src/util/tubemap.js
```diff
@@ -841,31 +887,66 @@ function compareNoNodeReadsByPreviousY(a, b) {
   const segmentA = reads[a.readIndex].path[a.pathIndex];
   const segmentB = reads[b.readIndex].path[b.pathIndex];
   if (segmentA.order === segmentB.order) {
-    return a.previousY - b.previousY;
+    // we want to sort of the reverse when the segment is between a cycle
+    // this ensures a loop that starts on the outside, stays on the outside
+    // and loops are rolled in order
```
adamnovak, line 892: This comment could be better written, I think. Maybe:

```suggestion
    // We want to sort in reverse order when the segment is along the reverse-going part of a cycle.
    // This ensures a loop that starts on the outside, stays on the outside,
    // and rolls up in order with other loops.
```

src/util/tubemap.js
```diff
@@ -676,12 +677,57 @@ function placeReads() {
...
+
+        // if the previous segment and the next segment is going to the same node
+        let nextNodeJustVisited = previousNodeOrder !== null && nextNodeOrder !== null && previousNodeOrder === nextNodeOrder;
+
+      
+        // a segment can also be in a loop if the next node is behind on the map
+        let nextNodeBehind = previousNodeOrder !== null && nextNodeOrder !== null && previousPathWithNode.order > nextNodeOrder;
+
+
+        reads[idx].path[pathIdx].betweenCycle = false;
```
adamnovak, line 711: I don't know if the idea of "being between a cycle" means anything to me. All the visits along the path that the read takes from a node back to the same node "participate in" the cycle, but here we're specifically interested in the visits where we're in the part of the cycle that goes from right to left when laid out, I think, right? Is there another way to say this?

src/util/tubemap.js
```diff
@@ -841,31 +887,66 @@ function compareNoNodeReadsByPreviousY(a, b) {
...
+  // The first index of each array contains the read index
+  // The second index of each array contains the path index
   let pathIndexA = a[1];
   let pathIndexB = b[1];
-  // let readA = reads[a[0]]
-  // let nodeIndexA = readA.path[pathIndexA].node;
-  let nodeA = nodes[reads[a[0]].path[pathIndexA].node];
-  let nodeB = nodes[reads[b[0]].path[pathIndexB].node];
+  let readIndexA = a[0];
+  let readIndexB = b[0];
```
adamnovak, line 910: Can't we use argument destructuring here?
```suggestion
function compareReadOutgoingSegmentsByGoingTo([readIndexA, pathIndexA], [readIndexB, pathIndexB]) {
```

adamnovak (changes requested):
This looks like it ought to work, but I think some parts of the code could be simplified and some of the comments could be made clearer.

adamnovak (approved):

PR #396, opened 2024-02-26
Fix Out of Order Veritcal Transitions

### LINKED ISSUE #395: Reads Vertical Transitions Out Of Order
Reads going from forward to reverse sometimes don't transition in the correct order.

Incorrect_Veritcal_Transition

Even though the horizontal segments are placed correctly, something is causing the vertical component to intersect.

ducku:
An offset designated for each node is used when generating vertical components of tracks. This leads to vertical transitions being determined by the order of the input tracks.

https://github.com/vgteam/sequenceTubeMap/blob/f4ab3c27dd53fc0ce0627d6a4885269378eb75ba/src/util/tubemap.js#L2942-L2953

https://github.com/vgteam/sequenceTubeMap/blob/f4ab3c27dd53fc0ce0627d6a4885269378eb75ba/src/util/tubemap.js#L3010

Tracks are now sorted based on how their horizontal components are ordered before generating vertical rectangles, allowing the offsets to work in a more predictable manner.

Correct_Veritcal_Transition

Closes #395

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -968,6 +968,37 @@ function compareReadIncomingSegmentsByComingFrom(a, b) {
   ); // neither has y-property
 }
 
+// Compare tracks based on ordering at their first convergence
+function compareTrackByInitialOrdering(trackA, trackB) {
+  // Find the first node where the two tracks converge, sort by layer
+  if (!trackA.hasOwnProperty("path")) return -1;
+  if (!trackB.hasOwnProperty("path")) return 1;
+
+  const pathA = trackA.path;
+  const pathB = trackB.path;
+
+  let AIndex = 0;
+  let BIndex = 0;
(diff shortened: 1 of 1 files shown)


src/util/tubemap.js
```diff
@@ -967,6 +967,37 @@ function compareReadIncomingSegmentsByComingFrom(a, b) {
...
+    } else if (trackAOrder < trackBOrder) {
+      AIndex += 1;
+    } else if (trackAOrder > trackBOrder) {
+      BIndex += 1;
+    } else {
+      // Orders are equal but are traversing out of nodes
+      AIndex += 1;
+      BIndex += 1;
+    }
+  }
```
adamnovak, line 995: This looks like it might get slow for a lot of long tracks needing to be compared, but I can't think of a better way to do it that's simple.

adamnovak (approved):

PR #399, opened 2024-03-06
Remove references to dataPath

### LINKED ISSUE #390: Get rid of the dataPath in the examples and tests
We no longer use the `dataPath` field on view targets, but it is still set on the examples in `config.js` and still checked in the link to view tests. We should get rid of it and only use the one global `dataPath`.

shreyasun:
Fixes #390

### DIFF
--- a/docker/config.json
+++ b/docker/config.json
@@ -7,7 +7,6 @@
         {"files": [{"type": "graph", "name": "snp1kg-BRCA1.vg.xg"}]},
         {"files": [{"type": "read", "name": "NA12878-BRCA1.sorted.gam"}]}
       ],
-      "dataPath": "default",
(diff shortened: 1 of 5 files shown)


adamnovak (changes requested):
It looks like the `dataPath` field giving the path to the exampleData folder itself got removed, when that usage needs to stay.

adamnovak:
Also it looks like commit [05cba44](https://github.com/vgteam/sequenceTubeMap/pull/399/commits/05cba44edf18b352eaaa8109364368956a6db6cc) actually *undid* removing `dataPath` from the truth link, so maybe it needs to be reverted?

adamnovak (approved):

PR #400, opened 2024-03-10
Extra Space Between Nodes

ducku:
Sometimes tracks don't have enough space to make a smooth transition from node to node. This PR creates extra space between nodes, taking into account the maximum difference in y values between nodes given any track. 

snp1kg-BRCA1
Before:
snp1kg-BRCA1_before

After:
snp1kg-BRCA1_after

chr1_15608265_15608765
Before:
chr1_15608265_15608765_before

After:
chr1_15608265_15608765_after

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -1937,7 +1937,7 @@ function calculateExtraSpace() {
           rightSideEdges[track.path[i].order] += 1;
         }
       } else {
-        // Track is going to a differnt node, account for space needed to limit rise/fall angle 
+        // Track is going to a different node; account for space needed to limit rise/fall angle 
         const yDifference = Math.abs(track.path[i].y - track.path[i - 1].y);
         //TODO: Extra space should also be accounted when there are too many tracks curving at the nodes
         fallAngleAdjustment[track.path[i].order] = Math.max(yDifference / 17.5, fallAngleAdjustment[track.path[i].order]);
@@ -3687,11 +3687,9 @@ function drawTrackRectangles(rectangles, type, groupTrack) {
     .text((d) => getPopUpTrackText(d.name));
 }
 
-function compareCurvesByLineChanges(a, b) {
+function compareCurvesByXYStartValue(a, b) {
   if (a.xStart < b.xStart) return 1;
   else if (a.xStart > b.xStart) return -1;
-  else if (a.xEnd < b.xEnd) return 1;
-  else if (a.xEnd > b.xEnd) return -1;
   else if (a.yStart > b.yStart) return 1;
   else if (a.yStart < b.yStart) return -1;
   return 0;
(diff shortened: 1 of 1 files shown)


adamnovak:
I like the space, but now it's easier to see that the reads are just sort of free-form swooshing around in these gaps, and their lines can get pretty thin along the axis perpendicular to their direction of motion. I wonder if there's a good way to use the extra space to ensure a minimum thickness or to get the reads to stick to horizontal and 45 degree angles.

ducku:
Currently, curves are drawn using SVG's Beziar Curves. We draw 2 parallel curves and fill them in.

Turns out, drawing [parallel Beziar Curves ](https://raphlinus.github.io/curves/2022/09/09/parallel-beziers.html)are pretty difficult. 

In commit 8b09fa2, some simple numbers were tweaked to make the curves look nicer.

In commit 14a46ec, I attempted to group sets of curves together and skew them based on how they're traversing. I think the results are better, but I'm not sure if it's the right way to do this. Maybe it's not a commit we want to keep.

snp1kg-BRCA1:
Skewed_Beziar_snp1kg-BRCA1

ch1_15608265_15608765:
Skewed_Beziar_ch1_15608265_15608765

A low hanging fruit could be adding more space between nodes based on the number of curvy reads. 

Some improvements can also be made to how curves are grouped. Sometimes, curves start and end on the same node, but aren't neccesarily traversing together. Other times, curves start and end on different nodes, but are traversing closely together. In snp1kg-BRCA1, there's a single read going from T that gets covered up by the reads going from C. Them being grouped together can prevent the coverup.

src/util/tubemap.js
```diff
@@ -1917,30 +1917,43 @@ function generateNodeXCoords() {
...
+      // Track is going to the same node, account for space taken up by edges looping around
       if (track.path[i].order === track.path[i - 1].order) {
         // repeat or translocation
         if (track.path[i].isForward === true) {
           leftSideEdges[track.path[i].order] += 1;
         } else {
           rightSideEdges[track.path[i].order] += 1;
         }
+      } else {
+        // Track is going to a differnt node, account for space needed to limit rise/fall angle
```
adamnovak, line 1940:
```suggestion
        // Track is going to a different node; account for space needed to limit rise/fall angle 
```

src/util/tubemap.js
```diff
@@ -3671,8 +3688,12 @@ function drawTrackRectangles(rectangles, type, groupTrack) {
 }
 
 function compareCurvesByLineChanges(a, b) {
```
adamnovak, line 3690: This no longer compares by the number of vertical lanes changed, so I think it probably needs a different name now,

adamnovak (changes requested):
It looks like the reads do narrow less now. This might be as good as we can get without doing fancy curve fitting to find parallel curves, or using some other curve representation.

I have a couple comments on the code that I think should be addressed, but otherwise I think this is good.

We might want to add a checkbox in the visualization options accordion to remove the new horizontal space if someone wants a very compact view.

adamnovak (approved):

PR #414, opened 2024-04-03
Remove Node Sequences Option

### LINKED ISSUE #349: Allow eliding node sequences from the server response
If we want to look at many megabases of sequence, with some kind of simplification, we don't necessarily want to include the node sequences in what we send to the client.

We need a flag in the API request to hide the sequences, and maybe a checkbox on the page (or in the graph track settings?) to control it.

shreyasun:
Fixes #349


adamnovak:
I think the test is failing because `&dataPath=default` got added back to the expected link during the merge.

src/components/HeaderForm.js
```diff
@@ -190,7 +194,7 @@ function viewTargetsEqual(currViewTarget, nextViewTarget) {
     return false;
   }
 
-  if (currViewTarget.simplify !== nextViewTarget.simplify) {
+  if (currViewTarget.removeSequences !== nextViewTarget.removeSequences) {
```
adamnovak, line 197: I think we also still need to check if the `simplify` fields differ, not *just* the new `removeSequences` field.

src/components/HeaderForm.js
```diff
@@ -825,6 +832,19 @@ class HeaderForm extends Component {
     this.setState({ simplify: !this.state.simplify });
   };
 
+  /* Function for toggling nodeSequences button, enabling client to make request for node sequences to be displayed or note */
```
adamnovak, line 835: I don't think `nodeSequences` is a variable or function name, and `note` is a typo. We could simplify the sentence to something like:

```suggestion
  /* Function for toggling display of node sequences */
```

src/components/HeaderForm.js
```diff
@@ -825,6 +832,19 @@ class HeaderForm extends Component {
...
+    this.setState({ removeSequences: !this.state.removeSequences });
+  };
+
+  openPopup = () => {
+    this.setState({ popupOpen: !this.state.popupOpen });
+  };
+
+  closePopup = () => {
+    this.setState({ popupOpen: !this.state.popupOpen });
+  };
```
adamnovak, line 846: These are called `openPopup` and `closePopup` but really they are two copies of a toggle-the-popup function. Either they should always do what their names say, or there should just be one `togglePopup`.

package.json
```diff
@@ -49,6 +49,7 @@
     "react-scripts": "5.0.1",
     "react-select": "^5.7.3",
     "react-select-event": "^5.5.1",
+    "react-switch": "^7.0.0",
```
adamnovak, line 52: This is not actually the switch component documented at https://mui.com/material-ui/react-switch/ that I thought we were using. This is a completely different package documented at https://www.npmjs.com/package/react-switch#usage which has about the same API.

This one might be a little better, so maybe we should keep it?

src/end-to-end.test.js
```diff
@@ -339,7 +339,7 @@ describe("When we wait for it to load", () => {
 it("produces correct link for view before & after go is pressed", async () => {
   // First test that after pressing go, the link reflects the dat form
   const expectedLinkBRCA1 =
-    "http://localhost?name=snp1kg-BRCA1&tracks[0][trackFile]=exampleData%2Finternal%2Fsnp1kg-BRCA1.vg.xg&tracks[0][trackType]=graph&tracks[0][trackColorSettings][mainPalette]=greys&tracks[0][trackColorSettings][auxPalette]=ygreys&tracks[1][trackFile]=exampleData%2Finternal%2FNA12878-BRCA1.sorted.gam&tracks[1][trackType]=read&region=17%3A1-100&bedFile=exampleData%2Finternal%2Fsnp1kg-BRCA1.bed&dataType=built-in&simplify=false";
+    "http://localhost?name=snp1kg-BRCA1&tracks[0][trackFile]=exampleData%2Finternal%2Fsnp1kg-BRCA1.vg.xg&tracks[0][trackType]=graph&tracks[0][trackColorSettings][mainPalette]=greys&tracks[0][trackColorSettings][auxPalette]=ygreys&tracks[1][trackFile]=exampleData%2Finternal%2FNA12878-BRCA1.sorted.gam&tracks[1][trackType]=read&dataPath=default&region=17%3A1-100&bedFile=exampleData%2Finternal%2Fsnp1kg-BRCA1.bed&dataType=built-in&simplify=false&removeSequences=false";
```
adamnovak, line 342: You probably don't mean to add back `dataPath`.

```suggestion
    "http://localhost?name=snp1kg-BRCA1&tracks[0][trackFile]=exampleData%2Finternal%2Fsnp1kg-BRCA1.vg.xg&tracks[0][trackType]=graph&tracks[0][trackColorSettings][mainPalette]=greys&tracks[0][trackColorSettings][auxPalette]=ygreys&tracks[1][trackFile]=exampleData%2Finternal%2FNA12878-BRCA1.sorted.gam&tracks[1][trackType]=read&region=17%3A1-100&bedFile=exampleData%2Finternal%2Fsnp1kg-BRCA1.bed&dataType=built-in&simplify=false&removeSequences=false";
```

src/server.mjs
```diff
@@ -382,6 +382,49 @@ api.post("/getChunkedData", (req, res, next) => {
...
 });
 
+
+/*
+graph = {
+  node: [
+    {
+      sequence: "AGCT"
+      id: "1"
+    ,
```
adamnovak, line 392:
```suggestion
    },
```

src/server.mjs
```diff
@@ -382,6 +382,49 @@ api.post("/getChunkedData", (req, res, next) => {
...
+  ]
+  edge: []
+  path: []
+}
+removing sequence would result in
+graph = {
+  node: [
+    {
+      id: "1"
+    ,
```
adamnovak, line 406:
```suggestion
    },
```

src/server.mjs
```diff
@@ -382,6 +382,49 @@ api.post("/getChunkedData", (req, res, next) => {
...
+      sequence: "AGCT"
+      id: "1"
+    ,
+    {
+      sequence: "AGCTAG"
+      id: "2"
+    }
+  ]
+  edge: []
+  path: []
```
adamnovak, line 399:
```suggestion
  ],
  edge: [],
  path: []
```

src/server.mjs
```diff
@@ -382,6 +382,49 @@ api.post("/getChunkedData", (req, res, next) => {
...
+  node: [
+    {
+      id: "1"
+    ,
+    {
+      id: "2"
+    }
+  ]
+  edge: []
+  path: []
```
adamnovak, line 412:
```suggestion
  ],
  edge: [],
  path: []
```

src/server.mjs
```diff
@@ -382,6 +382,49 @@ api.post("/getChunkedData", (req, res, next) => {
...
+*/
+
+// read a graph object and remove "sequence" fields in place
+function removeNodeSequencesInPlace(graph){
+  console.log("graph:", graph)
+  if (!graph.node){
+    return;
+  }
+  graph.node.forEach(function(node) {
+    console.log("node sequence to delete:", node)
```
adamnovak, line 423: We probably should cut this; people are going to remove sequence because there's a lot of it, and we don't want a lot of sequence going to the logs.
```suggestion
```

adamnovak (changes requested):
This looks like it should work, but there are some places where I would like the code cleaned up before merging it.

Also, for PRs that add new UI like this, it would be good to have a screenshot included to show how it looks.

src/components/HeaderForm.js
```diff
@@ -948,25 +968,45 @@ class HeaderForm extends Component {
...
+                  {!readsExist(this.state.tracks) && (
+                    <>
+                      <Button
+                        onClick={this.openPopup}
+                        outline
+                        active={this.state.simplify || this.state.removeSequences}
+                      >
+                      <FontAwesomeIcon icon={faGear} /> Simplify
+                      </Button>
+                      <PopupDialog open={this.state.popupOpen} close={!this.openPopup} width="400px">
```
adamnovak, line 987: I think `openPopup()` maybe needs to be renamed to `togglePopup()`? And you would just pass it here? It looks like you are trying to negate the function and I don't think this will actually work if you run it and try and open the dialog (so maybe that warrants a test?).

```suggestion
                      <PopupDialog open={this.state.popupOpen} close={this.togglePopup} width="400px">
```

src/components/HeaderForm.js
```diff
@@ -825,6 +836,15 @@ class HeaderForm extends Component {
     this.setState({ simplify: !this.state.simplify });
   };
 
+  /* Function for toggling display of node sequences */
+  toggleIncludeSequences = () => {
+    this.setState({ removeSequences: !this.state.removeSequences });
+  };
+
+  openPopup = () => {
```
adamnovak, line 844: This really toggles it.
```suggestion
  togglePopup = () => {
```

src/components/HeaderForm.js
```diff
@@ -948,25 +968,45 @@ class HeaderForm extends Component {
...
-                  {DataPositionFormRowComponent}
+                <div className="d-flex justify-content-between align-items-start">
+                  <div>
+                    {DataPositionFormRowComponent}
+                  </div>
+                  <div className="d-flex justify-content-end align-items-start flex-shrink-0">
+                  {!readsExist(this.state.tracks) && (
+                    <>
+                      <Button
+                        onClick={this.openPopup}
```
adamnovak, line 981: If we rename the function we'd change it here too.
```suggestion
                        onClick={this.togglePopup}
```

adamnovak (changes requested):
I don't think trying to use `!this.openPopup` as a close event handler is going to work. I think if you take the suggestions here the dialog will actually work when you go to open and close it manually; it looks like it is probably broken right now.

adamnovak (approved):

PR #425, opened 2024-04-30
Resolved extra horizontal space between nodes with a coordinate bar break

### LINKED ISSUE #405: Extra horizontal space should have a coordinate bar break
@ducku added extra horizontal space between nodes when there are lots of reads that need to move up and down. But this causes some large gaps/odd spacing in the horizontal axis tick marks that trace out node positions on the first/reference path.

They've always been not *perfectly* smooth and linear all the way across, but if we're going to make a big space on purpise we should indicate it on the axis somehow.

If we left space *out* it would be a normal axis break, but I'm not perfectly sure what to draw when we add space *in*.

shreyasun:
Fixes #405


adamnovak (approved):
This looks good to me!

PR #428, opened 2024-05-03
Add a script that can batch import and explain Lancet data in the November format

### LINKED ISSUE #427: Document which variant is meant to go with which Lancet example

adamnovak:
This fixes #427 sort of by making a BED that explains what variant is supposed to be seen.

I ran it like this:
```
cd ~/workspace/sequenceTubeMap
mkdir -p webhost
cd webhost

~/workspace/sequenceTubeMap/scripts/prepare_lancet_output.sh ~/Downloads/STM_DataShare_Nov07_2023\ 2/ ./lancet_2023-11-07

cd lancet_2023-11-07

python3 -m http.server
```

Then I could put `http://[::]:8000/index.bed` into my local tube map as the BED file and browse the data.

But, I think the data provided is not exactly what I really want to look at. It looks liek the called tumor variants are not actually *in* the graphs. For example, at `chr1:38506973-38506974` I am suppsoed to see a `Tumor-specific CTGGAATCCAGCAGCCCAGACTTCCACATCATAATTTTCTGGGGCAATGGTTTTCAAACTTCACTGTACG -> C DEL variant`. But the graph doesn't show a large deletion. Instead, that deleted sequence occurs 1 base into the leftmost node here, and I can see the softclips in the aligned tumor reads at their left ends, where they would read over the deletion edge that isn't present. Here's a screenshot of the softclips, with the tumor reads in red:

Screenshot 2024-05-03 at 16 56 40

So I now have a way to generate and host tumor-normal Lancet examples, including examples for larger indels, but the graphs coming from Lancet don't really seem to be the right ones.

### DIFF
--- a/scripts/prepare_lancet_output.sh
+++ b/scripts/prepare_lancet_output.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# prepare_lancet_input.sh: Turn a directory of Lancet output into a Tube Map BED file and directory that can be hosted on the web.
+set -x
+
+function usage() {
+    echo >&2 "${0}: Prepare a directory of Lancet data ."
+    echo >&2
+    echo >&2 "The input directory should have in it:"
+    echo >&2 "- variant_calling_regions.with_calls.bed, an extended BED file with biallelic VCF REF in column 6, ALT in column 7, and INFO in column 9 with a TYPE field"
+    echo >&2 "- giraffe_alignments/, a directory with, for each VCF record:"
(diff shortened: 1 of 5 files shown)


adamnovak:
OK, I talked to Rajeeva and he told me that Lancet makes several graphs per variant. I added code to find the most centered one for each variant, and now I can indeed see this variant. Here's that deletion being taken by just the tumor reads:

Screenshot 2024-05-07 at 17 24 37

These are now probably usable examples.

PR #435, opened 2024-05-15
Enable removing node sequences when reads are present

### LINKED ISSUE #422: Allow eliding node sequences when reads are present
We would like to be able to use node-sequence-removal simplification on Lancet graphs to see how tumor and normal reads flow through the graph structure.

shreyasun:
Fixes #422


adamnovak (approved):
This looks good to me.

It would be a better PR if it had screenshots, but I've seen it already.

PR #436, opened 2024-05-21
MinZoom

### LINKED ISSUE #434: You can hit the max zoom out in X without being able to see everything in Y
With node sequence removal and reads, a view can be much taller for its width than we expect, and we cna hit the max zoom out in X and not have the whole Y on screen. We should make the max zoom out make sure you can fit all of Y.

shreyasun:
Fixes #434


adamnovak (approved):
I had to drop some commented-out code but otherwise this is good.

PR #437, opened 2024-05-21
Aligned first and last coordinate

### LINKED ISSUE #429: First coordinate on the coordinate bar can start further left than I can pan

shreyasun:
Fixes #429

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -1399,8 +1399,9 @@ export function zoomBy(zoomFactor) {
 
   const minZoom = Math.min(
     1,
-    parentElement.clientWidth / (maxXCoordinate + 10)
-  );
+    parentElement.clientWidth / (maxXCoordinate + 10), 
+    parentElement.clientHeight / (maxYCoordinate + 10)
+  )
   const maxZoom = 8;
   const width = parentElement.clientWidth;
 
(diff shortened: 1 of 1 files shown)


src/util/tubemap.js
```diff
@@ -3711,14 +3728,25 @@ function drawRuler() {
 );
   
   // Plot all the ticks
-  ticks.forEach((tick) => drawRulerMarking(tick[0], tick[1]));
+  for (let i = 0; i < ticks.length; i++){
+    let tick = ticks[i];
+    let align;
```
adamnovak, line 3733:
```suggestion
    // Figure out how to align the tick text, to keep the outermost labels inside
    // the visible area
    let align;
```

src/util/tubemap.js
```diff
@@ -3711,14 +3728,25 @@ function drawRuler() {
...
+      align = "end";
+    } else {
+      align = "middle";
+    }
+    drawRulerMarking(tick[0], tick[1], align);
+  }
 }
 
-function drawRulerMarking(sequencePosition, xCoordinate) {
+function drawRulerMarking(sequencePosition, xCoordinate, align) {
```
adamnovak, line 3745:
```suggestion
/// Draw an axis tick for the given sequence position (in bp) at the given pixel X
/// coordinate. The text label can be aligned to the tick mark by its "start", "end",
/// or "middle".
function drawRulerMarking(sequencePosition, xCoordinate, align) {
```

adamnovak (approved):
This looks good. It could probably use some comments about *why* we need to change the alignment for the outer ticks though, to make it make sense why there's all this machinery.

PR #440, opened 2024-05-29
Modified sorting logic to sort no-node reads

### LINKED ISSUE #431: Reads from different samples interleave oddly when skipping nodes
See `Tumor-specific A -> AAAAAAATATATATATATAT INS variant at chr1:86645909` for example. The reads should stay separated by sample even as they skip nodes.

shreyasun:
Fixes #431

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -730,9 +730,7 @@ function placeReads() {
     });
   });
 
-  console.log("elementsWithoutNode before sort:", elementsWithoutNode)
   elementsWithoutNode.sort(compareNoNodeReads);
-  console.log("elementsWithoutNode:", elementsWithoutNode)
   elementsWithoutNode.forEach((element) => {
     const segment = reads[element.readIndex].path[element.pathIndex];
     segment.y = bottomY[segment.order];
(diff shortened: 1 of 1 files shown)


src/util/tubemap.js
```diff
@@ -874,20 +878,35 @@ function setOccupiedUntil(map, read, pathIndex, y, node) {
...
+  const segmentB = readB.path[b.pathIndex];
+  if (segmentA.order !== segmentB.order) {
+    return segmentA.order - segmentB.order;
+  } 
+  // Sort by reads' source track
+  const readA = reads[a.readIndex];
+  const readB = reads[b.readIndex];
+  if (readA.sourceTrackID !== readB.sourceTrackID){
+    return readA.sourceTrackID - readB.sourceTrackID;
+  }
```
adamnovak, line 893: The tests are failing because the reads are not defined early enough.

```suggestion
function compareNoNodeReads(a, b) {
  // Get the reads for the segments
  const readA = reads[a.readIndex];
  const readB = reads[b.readIndex];
  // Sort by order by segments
  const segmentA = readA.path[a.pathIndex];
  const segmentB = readB.path[b.pathIndex];
  if (segmentA.order !== segmentB.order) {
    return segmentA.order - segmentB.order;
  } 
  // Sort by reads' source track
  if (readA.sourceTrackID !== readB.sourceTrackID){
    return readA.sourceTrackID - readB.sourceTrackID;
  }
```

src/util/tubemap.js
```diff
@@ -723,12 +724,15 @@ function placeReads() {
...
+          previousNode: previousValidNode
         });
       }
     });
   });
 
-  elementsWithoutNode.sort(compareNoNodeReadsByPreviousY);
+  console.log("elementsWithoutNode before sort:", elementsWithoutNode)
+  elementsWithoutNode.sort(compareNoNodeReads);
+  console.log("elementsWithoutNode:", elementsWithoutNode)
```
adamnovak, line 735: I don't think we need this logging in the final version.
```suggestion
  elementsWithoutNode.sort(compareNoNodeReads);
```

src/util/tubemap.js
```diff
@@ -874,20 +878,35 @@ function setOccupiedUntil(map, read, pathIndex, y, node) {
 
 // compare read segments which are outside of nodes
 // by the y-coord of where they are coming from
```
adamnovak, line 880: This comment isn't true anymore and needs to be updated.

```suggestion
// Compare read segments which are outside of nodes, to sort them in a good horizontal
// and then vertical display order.
```

adamnovak (changes requested):
I think some of the code was reordered to add the comments and now it doesn't actually work. Also the comparison function's comment about what it does needs updating. I made some suggestions.

adamnovak (approved):

PR #456, opened 2024-08-20
Support for indexed GAF files

jmonlong:
The latest versions of vg support GAF indexing, so I added this support here too. I had an old branch with support that I updated. It seems to work 

## GAF support

- Use the appropriate command-line argument for `vg chunk`.
- Process the extracted GAF file by converting to GAM before piping to `vg view` to get the JSON representation.
- Added a test GAF file by converting, sorting and indexing `exampleData/cactus-NA12879.gam` (removing unmapped reads that tabix doesn't like...).

## Limitations

- Can't mix GAMs and GAFs because `vg chunk` can't handle it. Multiple GAMs **or** multiple GAFs work fine.

## Docker container update

- Bumped the version of vg to 1.59.0. 
- Updated the custom `config.json` for the container
   - Removed the examples bound to `dataPath` because the container is build so that `dataPath` points to `/data`


src/server.mjs
```diff
@@ -618,16 +618,35 @@ async function getChunkedData(req, res, next) {
...
-      // Use a GAM index
-      console.log("pushing gam file", gamFile);
-      vgChunkParams.push("-a", gamFile, "-g");
+      if (gamFile.endsWith(".gam")) {
+        // Use a GAM index
+        console.log("pushing gam file", gamFile);
+        anyGam = true;
+      }
+      if (gamFile.endsWith(".gaf.gz")) {
+        // Use a GAM index
```
adamnovak, line 636:
```suggestion
        // Use a GAF with index
```

PR #459, opened 2024-09-16
Make curves going the same way use a consistent Z order

adamnovak:
Right now crossing bands of reads will overlap arbitrarily. See the first region in the Lancet example (screenshot):
Screenshot of reads going up and reads going down overlapping each other randomly

With this change the reads going between a pair of nodes all have a consistent Z level:

Screenshot of all the reads going between one pair of nodes over all the reads going between a different pair

### DIFF
--- a/src/util/tubemap.js
+++ b/src/util/tubemap.js
@@ -4027,7 +4027,9 @@ function drawTrackCurves(type, groupTrack) {
 
   // Group track curves based on if they have the same start and end node
   myTrackCurves.forEach((curve) => {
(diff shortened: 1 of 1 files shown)


adamnovak:
We get a sort of transparency effect since the curve edges don't exactly coincide. We might want real transparency.

adamnovak:
This also seems to remove duplicate SVG nodes for curves, because we were drawing from the unfiltered `trackCurves` and not the filtered `myTrackCurves`.

